PR #7, opened 2017-06-17
Lower the requirement of R version

terrytangyuan:
This should allow a lot of old users to use this package. This would also help with our tensorflow/reticulate suit of packages that only requires >= 3.0.0.


lionel-:
I don't think that's possible because rlang requires 3.1. And even on that version there's still problems to fix (hopefully will be for the next version).

lionel-:
but requiring 3.4.0 is definitely an oversight, could you change to 3.1.0 please?

terrytangyuan:
I see. Done!

lionel-:
Thanks!

PR #11, opened 2017-06-27
Signal condition on empty selection

lionel-:
Includes #12 

With this PR the `recipes` package can issue an error on empty selections.

### DIFF
--- a/R/vars-select.R
+++ b/R/vars-select.R
@@ -166,6 +166,7 @@ vars_select <- function(.vars, ..., .include = character(), .exclude = character
 
   # Ensure all output .vars named
   if (is_empty(sel)) {
+    cnd_signal("tidyselect_empty", .mufflable = FALSE)
     names(sel) <- sel
   } else {
     unnamed <- names2(sel) == ""


lionel-:
https://github.com/tidyverse/tidyselect/pull/11/commits/8e6a2f0d2ccf28b20b4117e00eeba7664f329fd8 sends a condition when selection is empty

https://github.com/tidyverse/tidyselect/pull/11/commits/d164fd5764b9378735c1af05d51eefe90b3bce25 checks for old dplyr and overwrites the selecting function in the dplyr namespace so we don't get mismatch regarding the environment where variables are registered.

https://github.com/tidyverse/tidyselect/pull/11/commits/9d0f399937e8a83c071eb4cc69fe602dcfa53986 Renames all instances of `columns` to `variables` since the latter seems a bit more generic. This will probably break tests again though :/

hadley:
I think it would be better to keep compatible with dplyr by default, but add an argument to make it possible to change the name used in the error messages.

lionel-:
@hadley would it be reasonable to add an optional attribute to the `vars` object instead of an argument? This way select helpers like `one_of()` can have custom error messages as well.

hadley:
Sure

PR #12, opened 2017-07-10
Overwrite dplyr select functions in old versions

lionel-:
To avoid conflicts between the variables set by dplyr helpers and those set by tidyselect's.

### DIFF
--- a/R/tidyselect.R
+++ b/R/tidyselect.R
@@ -3,3 +3,59 @@
 #' @importFrom Rcpp cppFunction
 #' @useDynLib tidyselect, .registration = TRUE
 "_PACKAGE"
+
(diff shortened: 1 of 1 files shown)


R/tidyselect.R
```diff
@@ -3,3 +3,59 @@
...
+    current_vars = current_vars,
+    select_vars = vars_select,
+    rename_vars = vars_rename,
+    select_var = vars_pull
+  )
+
+  c(helpers_fns, ui_fns)
+}
+
+maybe_overtake_dplyr <- function(...) {
```
hadley: Maybe move this above `fns_table` just to provide the context?

PR #17, opened 2017-07-20
API for setting and getting vars

lionel-:
In the end I went for `replace()` and `query()` instead of `set()` and `get()` because I think it's better to restrict the latter terms for setting and getting elements from an object. It feels like the verbs for events, hooks, and stuff like tidyselect variables should be lexically distinct.


R/vars.R
```diff
@@ -0,0 +1,80 @@
...
+#' * `scoped_vars()` changes the current variables and sets up a
+#'   function exit hook that automatically restores the previous
+#'   variables once the current function returns.
+#'
+#' * `peek_vars()` returns the currently registered variables.
+#'
+#' @param vars A character vector of variable names.
+#' @param frame The frame environment where the exit hook for
+#'   restoring the old variables should be registered.
+#' @return For `poke_vars()` and `reset_vars()`, the old variables
```
hadley: Remove `reset_vars()`

R/vars.R
```diff
@@ -0,0 +1,80 @@
...
+#' # the state automatically when the function returns:
+#' fn <- function(vars) {
+#'   scoped_vars(vars)
+#'   one_of("d")
+#' }
+#' fn(letters)
+poke_vars <- function(vars) {
+  stopifnot(is_character(vars) || is_null(vars))
+
+  old <- vars_env$selected
```
hadley: I think this could be 

`env_set(selected, vars)`

Or does that require dev rlang?

R/vars.R
```diff
@@ -0,0 +1,80 @@
...
+
+  invisible(old)
+}
+#' @rdname poke_vars
+#' @export
+scoped_vars <- function(vars, frame = caller_env()) {
+  old <- poke_vars(vars)
+
+  # Inline everything so the call will succeed in any environment
+  expr <- lang(on.exit, lang(poke_vars, old), add = TRUE)
```
hadley: Why not 

```
expr(on.exit(poke_vars(!!old), add = TRUE))
``` 

?

R/vars.R
```diff
@@ -0,0 +1,80 @@
...
+  eval_bare(expr, frame)
+
+  invisible(old)
+}
+#' @rdname poke_vars
+#' @export
+peek_vars <- function() {
+  vars_env$selected %||% warn("Can't get tidyselect variables as none were registered")
```
hadley: Maybe just "No tidyselect variables were registered"?

R/vars.R
lionel-: Because this is evaluated in the calling environment. At the minimum it should be `expr(base::on.exit(tidyselect::poke_vars(!! old)))`, but maybe this will be evaluated in an environment where `::` is overscoped. In general I think it's best not to make assumptions about calling environments, either by using quosures or by inlining all symbols within the calls.

R/vars.R
lionel-: Good idea, but this is in dev rlang.

R/vars.R
hadley: Ah, I see. You could also write:

```R
expr(UQ(on.exit)(UQ(poke_vars)(!!old), add = TRUE))
```

right?

lionel-: yes that would be equivalent. I thought a clean way to do it would be a `lang_inline()` helper: https://github.com/tidyverse/rlang/issues/215

That would be less error-prone with complex expressions

PR #26, opened 2017-08-22
Revert to old behaviour for data expressions

lionel-:


NEWS.md
```diff
@@ -1,6 +1,34 @@
 
 # tidyselect 0.1.1.9000
 
+* The semantics for data expressions have been changed back to the old
+  behaviour because the new rules were causing a lot of trouble. From
+  now on data expressions (symbols and calls to `:` and `c()`) can
+  refer to both registered variables and to objects from the context.
+
+  However the semantics for context expressions (any calls other than
```
hadley: I thought we were changing this too.  What's the advantage of having special behaviour for `:` and `c`? Can't we tell people to unquote if they want to use variables in the overscope?

lionel-: That change was introduced in last dplyr version and it did not cause any issues. The rationale was that select helpers such as `starts_with()` should never hit anything from the overscope. I'm fine with reverting this as well, consistency is good and it's awkward having to explain that `select()` is different from the rest of the verbs.

hadley: Yeah, I think I'd prefer that

lionel-:
This now reverts everything back to dplyr 0.5.0 behaviour.

lionel-:
@hadley Now reverted back to dplyr 0.7.0 behaviour.

PR #27, opened 2017-08-22
Fix dplyr support

lionel-:
Includes #26


lionel-:
It seems the flag doesn't work because the namespace isn't populated yet.

We'll have to check for the dplyr version and make sure it stays consistent (no point releases).

PR #28, opened 2017-08-22
Add `.strict` argument to vars_select()

lionel-:
Closes tidyverse/dplyr#2981
Includes #26 

@hadley I'm not sure what's the best course here. Either we ignore:

- All errors, this seems like a bad idea.
- All unknown variable errors. This still seems too encompassing.
- All unknown variable errors but only in data expressions, as proposed in this PR.

Or is it better to not make the distinction with data expressions?


hadley:
I wonder if a better way of handling this would be to use object tables (with all undefined values returning `integer()`)

lionel-:
The issue with objectables is that we cannot know for sure that an object being looked up is not e.g. an S3 method. Returning `integer()` would cause trouble there.

Also the `exists()` method is not consistently called, R expects the `get()` method to return an unbound symbol. There are probably many issues with returning `integer()`.

I think the proper way is a patch for R that signals typed errors for unknown objects. If we ever get there, the code proposed in this PR can still be used for compatibility with old R versions.

lionel-:
Now rebased.

lionel-:
@hadley `.strict` has now the narrower meaning and cleaner implementation.

Also I rewrote `vars_rename()` to use tidy eval + that new approach. All expressions are evaluated in context-only. For consistency it now establishes a variable context, though I don't expect it will be terribly useful.

PR #30, opened 2017-08-25
Add last_var()

lionel-:
And rename `vars` argument of `one_of()` to `.vars`, which I forgot in last release.


lionel-:
By the way, maybe it's not too late to embed the helpers in the overscope since we can still export them in tidyselect and dplyr. Thanks to embedding tidyr will work better when dplyr is not attached (in packages many people call tidyr and dplyr with a namespace prefix). Also maintainers of packages that need embedded helpers  won't have to track additions like `last_var()`.

lionel-:
Now I'm thinking that `last_one()` may be a better name because it doesn't assumes what the variables are called (variable, column, predictor, ...).

hadley:
One is a bit too nonspecific. Maybe last col? I think that's most general of the other terms.

PR #38, opened 2017-09-06
`c()` handles strings like `:` and `-` - fixes #37

### LINKED ISSUE #37: string coercion affects `c()`
Was testing things out after reading Hadley's blog post, and found this edge case:

When using `c()`, we still get errors when combining string names and column numbers. Is it possible to work around this, as in `:`?

```r
library(tidyselect)

vars_select(names(mtcars), c(starts_with("m"), "cyl"))
#> Error: Strings must match column names. Unknown columns: 1


vars_select(names(mtcars), starts_with("m") : "cyl")
#>   mpg   cyl 
#> "mpg" "cyl" 
```

If you can't fix it, the blog post is a little inconsistent about whether this will work. The paragraph after the code block implies it will work for `c()`
> Luckily tidyselect 0.2.0 also introduces a few features that help writing safer code for data expressions. First, the support for strings and character vectors has been improved. All data expressions fully support strings. It is now valid to supply strings to - and ::
...
> Note that this only applies to c(), - or : because it would not make sense to write seq("name", "mass")

gergness:
Tried fixing it myself to see what it would look like - I can see why you're nervous about this @lionel- , the tests caught problems when a variable is named "c". 

I fixed by not adding the extra function to the overscope if there was a variable named c, but I think it could be improved.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 
 # tidyselect 0.2.0.9000
+* `c()` now supports character vectors the same way as `-` and `seq()`.
+  (#37 @gergness)
 
 * `-` now supports character vectors in addition to strings. This
   makes it easy to unquote column names to exclude from the set:
--- a/R/vars-select.R
+++ b/R/vars-select.R
@@ -217,6 +217,9 @@ vars_select_eval <- function(vars, quos) {
 
   # Overscope `:` and `-` with versions that handle strings
   data <- c(data, `:` = vars_colon, `-` = vars_minus)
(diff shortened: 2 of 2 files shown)


lionel-:
You will have to create a new overscope where the data has precedence over the functions. Then you can call `overscope_eval_next()`. This should also be a bit more efficient since the overscope will be created only once.

lionel-:
Thanks for looking into this by the way. If you don't have time to delve into the overscope stuff I can take over from here no worries.

gergness:
Yea, that'd be great, thank you!

PR #60, opened 2018-04-16
typos in select helpers

dpprdan:
- "`ends_with()`: ends with a suffix." (not "with a prefix")
- add "." where missing


lionel-:
Thanks!

lionel-:
Hmm the bullet list is a mix of full and non-full sentences. I rewrote all items as sentences with a full stop.

PR #95, opened 2019-06-24
doc: integrate description of .strict in vars_select and vars_rename

atusy:
vars-rename.R and vars-select.R both describes the `.strict` parameter, and thus the document has description of `.strict` twice.
This PR fixes it.

Note that I did not run `devtools::document` because I use roxygen 6.1.1 whereas the current version of `tiyselect` uses 6.1.0, and also because I got `Error in getDLLRegisteredRoutines.DLLInfo(dll, addNames = FALSE)`

https://github.com/tidyverse/tidyselect/blob/b8c83cc0292e04701bf0a1b892da4bb1aee47ed3/R/vars-rename.R#L3-L4
https://github.com/tidyverse/tidyselect/blob/b8c83cc0292e04701bf0a1b892da4bb1aee47ed3/R/vars-select.R#L32

### DIFF
--- a/R/vars-rename.R
+++ b/R/vars-rename.R
@@ -1,7 +1,5 @@
 #' @export
 #' @rdname vars_select
-#' @param .strict If `TRUE`, will throw an error if you attempt to rename a
-#'   variable that doesn't exist.
 vars_rename <- function(.vars, ..., .strict = TRUE) {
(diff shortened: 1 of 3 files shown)


lionel-:
Thanks!

PR #97, opened 2019-06-24
Bump minimum R version to 3.2

atusy:
This PR fixe failure in Travis CI.

Tidyverse packages are going to support only R 3.2 or greater, and the suggested dplyr package on CRAN is already doing so.
Thus, I bumped minimum R version to 3.2 in DESCRIPTION and .travis.yml.


lionel-:
Thanks!

PR #101, opened 2019-07-08
Add better error message for tidyselect helpers

### LINKED ISSUE #73: Better error message in tidyselect helpers
> No tidyselect variables were registered.

Should probably be something like:

```
Error: `one_of()` must be used within a *selecting* function
```

zkamvar:
This will fix #73 when merged.

I found that the message "No tidyselect variables were registered." came from `peek_vars()`, which triggered this when there were (understandably) no vars registered in the `vars_env`. 

This now throws better errors like so:

``` r
library("tidyselect")
starts_with("z")
#> `starts_with()` must be used within a *selecting* function
peek_vars()
#> No tidyselect variables were registered.
```

<sup>Created on 2019-07-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Note that I am unsure of where an option to override the error would go: in the `peek_vars()` or in the helpers?

### DIFF
--- a/R/vars.R
+++ b/R/vars.R
@@ -80,8 +80,14 @@ poke_vars <- function(vars) {
 }
 #' @rdname poke_vars
 #' @export
-peek_vars <- function() {
-  vars_env$selected %||% abort("No tidyselect variables were registered")
+peek_vars <- function(generic_error = FALSE) {
+  if (is.null(vars_env$selected)) {
+    the_call <- sys.call(sys.parent()) 
+    fun <- as.character(the_call)[1]
+    msg <- sprintf("`%s()` must be used within a *selecting* function", fun)
+    abort(msg)
+  } 
+  vars_env$selected
 }
 
 #' @rdname poke_vars
--- a/R/vars.R
+++ b/R/vars.R
@@ -79,15 +79,22 @@ poke_vars <- function(vars) {
   invisible(old)
 }
 #' @rdname poke_vars
+#' @param fn the name of the function to use in the error message. Defaults to
+#'   `NULL`, which will use the first element of `sys.call(sys.parent())`.
 #' @export
-peek_vars <- function() {
+peek_vars <- function(fn = NULL) {
   if (is.null(vars_env$selected)) {
-    the_call <- sys.call(sys.parent()) 
-    fun <- as.character(the_call)[1]
-    if (fun == "peek_vars") {
-      msg <- "No tidyselect variables were registered."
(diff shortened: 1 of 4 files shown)


zkamvar:
As an aside, since this was not specifically identified in #73, I'm wondering about lines like this:

https://github.com/tidyverse/tidyselect/blob/f90cbe90adda0516ca54fc59c5c1c989cdd3f888/R/select-helpers.R#L51-L52

This gives the user `nchar(match) > 0 is not TRUE`, and I think it may be better just to write a generic error handler for these as opposed to relying on `stopifnot()`. Should I attempt to make changes in this PR?

R/vars.R
```diff
@@ -81,7 +81,17 @@ poke_vars <- function(vars) {
 #' @rdname poke_vars
 #' @export
 peek_vars <- function() {
-  vars_env$selected %||% abort("No tidyselect variables were registered")
+  if (is.null(vars_env$selected)) {
+    the_call <- sys.call(sys.parent()) 
+    fun <- as.character(the_call)[1]
```
lionel-: I think this would be a bit more flexible if `peek_vars()` took `fn = NULL` parameter. Then the caller could override the function name.

Also for robustness it would be nice to check if the call head (e.g. `the_call[[1]]` is a symbol. If it is not, we'd fall back to the default error message.

zkamvar: Thank you! I was having trouble conceiving of what the argument (`fn = NULL`) should be. 

For the robustness aspect, something like this?

```r
if (is.symbol(the_call[[1]]) && !identical(the_call[[1]], as.symbol('peek_vars'))) {
  # call-specific
} else {
  abort("No tidyselect variables were registered")
}
```

lionel-: yup though I'm not sure we need to have a special-case for `peek_vars()`?

zkamvar: If we didn't have the special case for peek_vars(), then the user would get the error "peek_vars() must be used within a *selecting* function." if they attempted to use it in the Global Environment, which doesn't make too much sense, since peek_vars() is not used in the selecting functions.

zkamvar: Although, it does create its own host of issues for testing -_-

lionel-: hmm, though that's not entirely wrong either :) For simplicity I would just omit the special-casing.

The new `fn` parameter looks good!

zkamvar: gotcha! Will fix.

lionel-:
Thank you!

Issue #96, opened 2019-06-24
Add pkgdown site


batpigandme:
This will allow for internal linking from other pkgdown sites (e.g. from dplyr to select helpers). Also could be useful place to have more detail à la @lionel-'s talk at rstudio::conf(2019).

hadley:
Just added the dns settings so http://tidyselect.r-lib.org should work shortly

PR #102, opened 2019-07-08
add pkgdown site

statnmap:
issue #96

Verify url to github if moving from tidyverse to r-lib.  
Also need to be addressed a valid URL as output.  
On my laptop, this looks like this:  
tidyselect

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -6,3 +6,5 @@
 ^revdep$
 ^codecov\.yml$
 ^\.github$
+^_pkgdown\.yml$
+^docs$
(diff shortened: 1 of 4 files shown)


statnmap:
Now you can try to implement it. `starts_with` should be protected.  
I also changed URL of github to use `r-lib` instead of `tidyverse`

lionel-:
Thanks!

statnmap:
You may add the URL in the description of the github repo as soon as Travis has built it: https://tidyselect.r-lib.org/

lionel-:
hmm I think I first need to add some environment variables to the Travis settings for the repo

Issue #92, opened 2019-04-19
rename vs names()<-  performance


grayskripko:
``` r
library(tidyverse)

n_cols <- 5000

one_row_test_tbl <- 1:n_cols %>%
  set_names(str_c('some_long_prefix', 1:n_cols)) %>%
  as.list() %>% 
  as_tibble()

system.time(tidyverse_renamed <- one_row_test_tbl %>% rename_all(toupper))
#>    user  system elapsed 
#>   20.34    0.00   20.36

system.time({
  base_renamed <- one_row_test_tbl
  names(base_renamed) <- one_row_test_tbl %>% names() %>% toupper()
})
#>    user  system elapsed 
#>       0       0       0

stopifnot(setequal(names(tidyverse_renamed), names(base_renamed)))
```

<sup>Created on 2019-04-19 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Is this difference in performance expected?

lionel-:
Calling `rlang::as_data_mask()` on `data` before evaluation should improve performance.

PR #104, opened 2019-07-08
add data mask to vars_rename_eval

zkamvar:
This will address #92, but I'm not sure how to add performance regression testing ᕕ( ᐛ )ᕗ

### DIFF
--- a/R/vars-rename.R
+++ b/R/vars-rename.R
@@ -43,7 +43,7 @@ vars_rename_eval <- function(quos, vars) {
   # Only symbols have data in scope
   is_symbol <- map_lgl(quos, is_symbol_expr)
   data <- set_names(as.list(seq_along(vars)), vars)
-  mask <- rlang::as_data_mask(data)
+  mask <- as_data_mask(data)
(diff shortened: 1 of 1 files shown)


R/vars-rename.R
```diff
@@ -43,7 +43,8 @@ vars_rename_eval <- function(quos, vars) {
   # Only symbols have data in scope
   is_symbol <- map_lgl(quos, is_symbol_expr)
   data <- set_names(as.list(seq_along(vars)), vars)
-  renamed <- map_if(quos, is_symbol, eval_tidy, data)
+  mask <- rlang::as_data_mask(data)
```
lionel-:
```suggestion
  mask <- as_data_mask(data)
```

lionel-:
Thank you!

Unfortunately we don't have a good setup for testing performance regressions. Your patch definitely solves the problem though:

```r
# Before
system.time(dplyr::rename_all(df, identity))
#>    user  system elapsed
#>  31.825   0.125  32.032

# After
system.time(dplyr::rename_all(df, identity))
#>    user  system elapsed
#>   0.282   0.000   0.283
```

PR #107, opened 2019-08-12
Add pkgdown site to DESCRIPTION URL

batpigandme:
This should make it possible for us to link directly to the tidyselect site from other package documentation (e.g. see https://github.com/tidyverse/dplyr/issues/4529).


lionel-:
Thanks!

PR #115, opened 2019-09-11
Warn in case of duplicate renaming

### LINKED ISSUE #52: Warning if names of duplicated selected variables do not agree
I get the idea of having select deduplicate column names/indices, but wouldn't some form of warning make sense? Here one particularly quirky example: the position of the column seems to be determined by the first appearance of the duplicated index, but the name by the last appearance.

(First mentioned in https://github.com/tidyverse/dplyr/issues/2354)

```R
tibble(x="A",y="B") %>%
  select(y=2, x=1, z=2)
# A tibble: 1 x 2
  z     x    
  <chr> <chr>
1 B     A 
```

lionel-:
Closes #52.

### DIFF
--- a/R/vars-select.R
+++ b/R/vars-select.R
@@ -253,11 +253,28 @@ inds_unique <- function(x) {
 
   # Keep last name of duplicates
   if (length(out) < length(x)) {
-    names <- map(split$val, names)
-    names(out) <- map_chr(names, last)
+    names(out) <- map_chr(split$val, ind_last_name)
   }
 
   out
+}
+
+ind_last_name <- function(x) {
+  names <- names(x)
+  last <- last(names)
+
+  if (!all(vctrs::vec_duplicate_detect(names) | names == "")) {
+    dups <- encodeString(names, quote = "\"")
+    dest <- last(dups)
+    dups <- glue::glue_collapse(dups, ", ", last = " and ")
+    msg <- glue::glue(
+      "Can't rename the same variable to different names ({dups}).
+       The variable will be renamed to the last one ({dest})."
+    )
+    warn(msg, "tidyselect_warning_duplicate_renaming", var = x)
+  }
+
+  last
 }
 
 ind_check <- function(x) {
--- a/R/vars-select.R
+++ b/R/vars-select.R
@@ -236,7 +236,7 @@ inds_combine <- function(vars, inds) {
     incl <- seq_along(vars)
   } else {
     incl <- inds[inds > 0]
-    incl <- inds_unique(incl)
+    incl <- inds_unique(incl, vars)
   }
 
   # Remove variables to be excluded (setdiff loses names)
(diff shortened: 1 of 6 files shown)


R/vars-select.R
```diff
@@ -229,20 +243,36 @@ inds_combine <- function(vars, inds) {
...
 
   out
 }
 
+ind_last_name <- function(x) {
+  names <- names(x)
+  last <- last(names)
+
+  if (!all(vctrs::vec_duplicate_detect(names) | names == "")) {
+    dups <- encodeString(names, quote = "\"")
```
hadley: I think this should be `` ` ``? (Since it's a variable name not a string)

R/vars-select.R
```diff
@@ -229,20 +243,36 @@ inds_combine <- function(vars, inds) {
...
+    dest <- last(dups)
+    dups <- glue::glue_collapse(dups, ", ", last = " and ")
+    msg <- glue::glue(
+      "Can't rename the same variable to different names ({dups}).
```
hadley: Can we get the name of the variable being renamed here? Then the message could be:

```
Must rename a variable to a single choice: {var} is being renamed to {dups}. 
Using `{dest} = {var}`
```

R/vars-select.R
lionel-: oh of course

R/vars-select.R
lionel-: I like it. Maybe this would be even clearer this way?

```
Can't rename variables to multiple choices: `a` is being renamed to `d` and `e`.
Renaming to `e`. 
```

hadley: The style guide suggests using must when we know what the correct result is, but I agree that my wording still seems clumsy. Maybe @batpigandme will have other ideas.

batpigandme: I like the call out of `Can't rename variables to multiple choices` (it's clear what the action was that is forbidden), and/but `Must rename a variable to a single choice` is the corrective. 

Is the behaviour that it automatically renames to the last name given? If not, you could use the `Must rename variable to a single choice using {dest} = {var}`.

lionel-: I don't like "Must" here because I think it implies there should be an error rather than a warning. Here we "can't" rename to multiple choices, but we have taken action to continue nonetheless.

lionel-: Also how about

> Renaming `a` to `e`.

instead of

> Using `e = a`.

The latter will make users pause as they decipher the new = old syntax.

batpigandme: Ah, your points are noted and make sense to me.

lionel-: WDYT of my last two comments @hadley?

lionel-: I no longer think that "must" is a problem for warnings. The message is now:

```
Must rename variables to a single choice: `a` is being renamed to `d` and `e`.
Renaming `a` to `e` instead.
```

However, I still find this message structure confusing. When we know both what can't happen and what must be done, I think I'd prefer:

```
Can't rename `a` to `d` and `e` because variables must be renamed to a single choice.
Renaming `a` to `e` instead.
```

I find it clearer because we start with the action that the user attempted rather than with what they should do instead. I think it makes it easier for the user to connect the warning to their code.

lionel-:
Merging because I need this to go forward with #118, but please let me know of any further comments.

PR #116, opened 2019-09-11
Don't return duplicate names after renaming variables

### LINKED ISSUE #70: Tibble allows duplicate column names after rename().
*[move[bot]](https://github.com/apps/move) commented on Jul 21, 2018, 9:58 PM UTC:*

_@kovla commented on Jul 12, 2018, 9:35 AM UTC:_

I imagine that `rename()` of a column to a name that already exists in the tibble (in another column) should warrant at least a warning. Otherwise it can result in two indistinguishable variables, with same name and type:

```
x <- tibble(a = 1:3, b = 4:6, c = 7:9)
x %>% rename(b = c)

# A tibble: 3 x 3
      a     b     b
  <int> <int> <int>
1     1     4     7
2     2     5     8
3     3     6     9

```

_This issue was moved by [krlmlr](https://github.com/krlmlr) from [tidyverse/tibble#431](https://github.com/tidyverse/tibble/issues/431)._

*This issue was moved by [romainfrancois](https://github.com/romainfrancois) from [tidyverse/dplyr#3723](https://github.com/tidyverse/dplyr/issues/3723).*

lionel-:
Closes #70

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,6 +58,9 @@
 * `vars_rename()` now fails when a variable is renamed to an existing
   name (#70).
 
+* `vars_rename()` has better support for existing duplicates (but
+  creating new duplicates is an error).
+
 * `vars_select()`, `vars_rename()` and `vars_pull()` now detect
   missing values uniformly (#72).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,6 +55,9 @@
 * `vars_rename()` no longer ignore the names of unquoted character
   vectors of length 1 (#79).
 
+* `vars_rename()` no longer creates duplicates when a variable is
+  renamed to an existing name (#70).
+
 * `vars_select()`, `vars_rename()` and `vars_pull()` now detect
   missing values uniformly (#72).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,8 +55,8 @@
 * `vars_rename()` no longer ignore the names of unquoted character
   vectors of length 1 (#79).
 
-* `vars_rename()` no longer creates duplicates when a variable is
-  renamed to an existing name (#70).
+* `vars_rename()` now fails when a variable is renamed to an existing
+  name (#70).
 
 * `vars_select()`, `vars_rename()` and `vars_pull()` now detect
   missing values uniformly (#72).
--- a/R/utils.R
+++ b/R/utils.R
@@ -109,3 +109,7 @@ last <- function(x) {
 str_compact <- function(x) {
   x[x != ""]
 }
+
+vec_index_is_empty <- function(x) {
+  !length(x) || all(x == 0L)
+}
--- a/R/utils.R
+++ b/R/utils.R
@@ -109,7 +109,3 @@ last <- function(x) {
 str_compact <- function(x) {
   x[x != ""]
 }
-
-vec_index_is_empty <- function(x) {
-  !length(x) || all(x == 0L)
-}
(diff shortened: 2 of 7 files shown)


hadley:
@jennybc does this seem consistent with existing (re)naming behaviour elsewhere

lionel-:
This seems consistent with `mutate()`, `$<-`, etc.

jennybc:
With the caveat that I feel like I'm looking at this with no context, I don't see a problem. So, it's clear you want to reduce to the remaining unique names, as opposed to uniquify-ing them using the `...j` strategy? I guess I don't understand if you need to preserve length, but I gather from this solution that you do not.

lionel-:
I don't think we have any size constraints that would prevent us from making names unique. I ruled this solution out because I thought it would be weird if the data frame ended up without the name explicitly chosen by the user. Unless we only change the pre-existing name? This would have the consequences:

- A message is printed, which reminds the user about the existing variable in case they have forgotten.

- The column is kept around and can be inspected.

We could have `.homonyms = c("keep", "drop," "error")`. It might be annoying to add a step to the process of renaming to an existing variable though. In R the user is usually able to overwrite columns or elements as part of the normal workflow.

jennybc:
I just read the original issue (#70). Now that I understand better, it does seem self-inconsistent that `dplyr::rename()` will yield non-unique column names, without, e.g. explicitly permitting them via `.name_repair = "minimal"`.

lionel-:
Current behaviour:

```r
tibble(x = 1, y = 2) %>% rename(x = y)
#> # A tibble: 1 x 2
#>       x     x
#>   <dbl> <dbl>
#> 1     1     2
```

Proposed behaviour (in this PR):

```r
tibble(x = 1, y = 2) %>% rename(x = y)
#> # A tibble: 1 x 1
#>       x
#>   <dbl>
#> 1     2
```

Alternatively, two ways of making names unique:

```r
tibble(x = 1, y = 2) %>% rename(x = y)
#> New names:
#> * x -> x...1
#> # A tibble: 1 x 2
#>   x...1     x
#>   <dbl> <dbl>
#> 1     1     2

tibble(x = 1, y = 2) %>% rename(x = y)
#> New names:
#> * x -> x...1
#> * x -> x...2
#> # A tibble: 1 x 2
#>   x...1 x...2
#>   <dbl> <dbl>
#> 1     1     2
```

lionel-:
If we go with make-unique, the user can fix their code with:

```r
tibble(x = 1, y = 2) %>% mutate(x = NULL) %>% rename(x = y)

tibble(x = 1, y = 2) %>% mutate(x = y, y = NULL)
```

hadley:
I think the key question is: Does `rename(df, y = x)` clearly indicate that you want to *replace* an existing variable called `y`, or is it an accident? i.e. Is `rename(df, y = x)` more like `df$y <- df$x` or more like `tibble(y = 1:5, y = 5:1)`?

jennybc:
I definitely think of `rename()` as a request to edit a name, so this:

Screen Shot 2019-09-13 at 7 38 41 AM

lionel-:
That makes sense.

hadley:
I think @jennybc's metaphor is compelling, and indicates that we should treat renaming to replace an existing variable as an accident, and hence we need `vec_as_names()` handling.

lionel-:
`vars_rename()` has been refactored to share code paths with `vars_select()`. It now ignores existing duplicates and fails when new ones are created.

PR #117, opened 2019-09-11
Support S3 vectors in `vars_select()` and `one_of()`

### LINKED ISSUE #109: select() should warn on factor arguments
The guide on tidy evaluation suggests [forwarding `...` to dplyr functions](https://tidyeval.tidyverse.org/dplyr.html#patterns-for-multiple-arguments) to handle multiple arguments in an NSE setting. In general, this works extremely well and allows custom functions to take advantage of the dplyr flexibility. For example,

``` r
from_mtcars <- function(...) {mtcars %>% select(...)}
from_mtcars( vs, am )              # Automatic quoting of arguments
from_mtcars( starts_with("d") )    # Select helper support
```

Unfortunately, almost invariably somebody passes a factor to the custom function, which leads to unexpected behavior. This usually happens when the user is working with traditional data frames and forgets to do `stringsAsFactors=FALSE`:

``` r
df <- data.frame( v=c("vs","am") )
from_mtcars( df$v )
#                     cyl  mpg
# Mazda RX4             6 21.0
# Mazda RX4 Wag         6 21.0
# ...
```

The behavior is due to factors getting converted to integer indices. This creates hard-to-catch bugs in the end user code, because the conversion is silent and the user usually expects the input to behave as strings. The author of the custom function can add a guard against this:

``` r
from_mtcars2 <- function(...)
{
    v <- purrr::map_if( list(...), is.factor, as.character )
    mtcars %>% select(!!!v)
}
```

Unfortunately, this breaks NSE support:
``` r
# Works as expected with factors
from_mtcars2( factor(c("vs","am")) )
#                     vs am
# Mazda RX4            0  1
# Mazda RX4 Wag        0  1
# ...

# But not with unevaluated expressions
from_mtcars2( mpg )
# Error in map_lgl(.x, .p, ...) : object 'mpg' not found
```

Since `select()` already has low-level access to variables, I was hoping that it could automatically check for the presence of factor arguments. If nothing else, I think that throwing a warning upon encountering one can help alert end users that they are about to encounter unexpected behavior. Potential place where this could be done is around

https://github.com/r-lib/tidyselect/blob/4de95ee641b68f7ab098816d01e02737afa2410c/R/vars-select.R#L144

where character arguments get mapped to integer positions.

lionel-:
Closes #109.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
 
 # tidyselect 0.2.5.9000
 
-* `vars_select()` now supports unquoting factors and other S3 vectors.
+* `one_of()` now supports factors and other S3 vectors. Similary,
+  `vars_select()` now supports unquoting S3 vectors.
 
 * `vars_rename()` no longer ignore the names of unquoted character
   vectors of length 1 (#79).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 
 # tidyselect 0.2.5.9000
 
-* `one_of()` now supports factors and other S3 vectors. Similary,
+* `one_of()` now always coerces its input to a character, or fails. Similary,
   `vars_select()` now supports unquoting S3 vectors.
 
 * `vars_rename()` no longer ignore the names of unquoted character
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 
 # tidyselect 0.2.5.9000
 
+* `vars_select()` now supports unquoting factors and other S3 vectors.
+
 * `vars_rename()` no longer ignore the names of unquoted character
   vectors of length 1 (#79).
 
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 # tidyselect 0.2.5.9000
 
+* `one_of()` now supports factors and other S3 vectors. Similary,
```
hadley:
```suggestion
* `one_of()` now always coerces it is input to a character, or fails. Similary,
```

R/select-helpers.R
```diff
@@ -113,11 +113,13 @@ num_range <- function(prefix, range, width = NULL, vars = peek_vars()) {
 one_of <- function(..., .vars = peek_vars()) {
   keep <- list(...)
 
-  bad_input <- detect_index(keep, negate(is.character))
+  bad_input <- detect_index(keep, ~ !vec_is_coercible(., chr()))
```
hadley: I'd add a comment here cross-referencing the vctrs issue.

Issue #89, opened 2019-03-31
Tidyselect functions should take a prototype as input


hadley:
Rather than a vector of variable names, as this would allow tidy select to also support type predicates.

(We'd need to transition to this gradually - probably first by adding a `ptypes` argument, then by deprecating vars, then removing vars)

lionel-:
The problem with this approach is that tidyselect takes the base R vector view, not the vctrs vector view. If you want to select elements from a list, the prototype will contain no information about the names.

hadley:
Where do we select elements from a list? That's definitely a rare case, so I don't think we should optimise for that (but maybe you're just arguing we shouldn't deprecate the old interface?)

lionel-:
I think it's strange to take a prototype approach which is a vctrs concept and then use `names()` which is a base vector concept.

> but maybe you're just arguing we shouldn't deprecate the old interface

Yes I think so. Maybe adding `df_select()` and `df_select_vars()` makes sense? The former would return a data frame and the latter just names. If we ever want to allow actions within selections, the former could put the data mask in dynamic scope. Then we could have UIs like

```r
facet_grid(rows = starts_with("foo"), cols = acts(cut_number(foo, 3)))
```

hadley:
We'd still returning a vector of names in either case. This is about the input interface.

lionel-:
Yes as in `df_select_vars()`. But `df_select()` with input + output df might also make sense if we need to allow actions in selections.

hadley:
When we make this new interface, I think we should also stop taking `...` and instead take a single `cols` argument.

hadley:
This interface would have also prevented https://github.com/tidyverse/tidyr/pull/668

lionel-:
Needs to return a named vector of variables instead of a prototype, as this is the only way to support on-the-fly renaming:

```r
mtcars %>% select(foo = cyl)
```

lionel-:
Instead of supporting functions in selections (issue #86), it may be better to have a special kind of helper to inspect the prototypes? It might be clearer for users that the values cannot be inspected:

```r
data %>% select(types(is.character) | types(is.factor))

data %>% mutate(mapping(types(is.numeric)), standardise))
```

versus

```r
data %>% select(is.character | is.factor)

data %>% mutate(mapping(is.numeric, standardise))
```

hadley:
[...]

PR #118, opened 2019-09-18
Support duplicate variables in `vars_select()`

### LINKED ISSUE #94: select should be able to drop columns by number when duplicate names exist
When a data.frame contains columns with the same name, select can't be used to correct the problem by dropping columns.

Even if rename and other verbs are problematic in this regard, dropping duplicate columns should (in principle) be possible.

---

Consider these error cases:

```r
tibble::tribble(
  ~a, ~a, ~b,
  1, 2, 3
) %>% 
  select(-1)
```

or

```r
tibble::tribble(
  ~a, ~a, ~b,
  1, 2, 3
) %>% 
  select(starts_with("a"))
```

lionel-:
Closes #94.

`vars_select()` now ignores duplicate variables that were not selected:

```r
vars <- c("a", "b", "a")

vars_select(vars, -3)
#>   a   b
#> "a" "b"

vars_select(vars, b)
#>   b
#> "b"
```

If selected, the duplicate variables are renamed to their repaired form:

```r
vars_select(vars, a)
#> New names:
#> * a -> a...1
#> * a -> a...2
#> a...1 a...2
#>   "a"   "a"

vars_select(vars, one_of(c("b", "a")))
#> New names:
#> * a -> a...2
#> * a -> a...3
#>     b a...2 a...3
#>   "b"   "a"   "a"
```

This repairing is not very useful yet because `vars_select()` returns a character vector. Ideally it'd return an integer vector of positions so the duplicate variables can be disambiguated. We'll tackle this as part of the UI update in #89.

I don't expect many breakages as duplicate variable names are a rare corner case. As far as `dplyr::select()` is concerned, this will continue to be an error because it checks for unique values:

```r
dups <- new_tibble(list(a = 1, b = 2, a = 3), nrow = 1)

dplyr::select(dups, b, a)
#> New names:
#> * a -> a...2
#> * a -> a...3
#> Error: Column `a` must have a unique name
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,9 +43,7 @@
   only the last one is taken into account (#52). The warning inherits
   from `tidyselect_warning_duplicate_renaming`.
 
-* `vars_select()` now supports duplicate variables (#94). They are
-  ignored if not part of the selection. Otherwise, they are repaired
-  to make them unique.
+* `vars_select()` now ignores existing duplicate variables (#94).
 
 * `one_of()` now always coerces its input to a character, or fails. Similary,
   `vars_select()` now supports unquoting S3 vectors.
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,7 +45,9 @@
 
 * `vars_select()` now ignores existing duplicate variables (#94). It
   fails if user attempts to rename to an existing variable, or if
-  different variables are renamed to the same name.
+  different variables are renamed to the same name. The errors inherit
+  from `tidyselect_error_rename_to_existing` or
+  `tidyselect_error_rename_to_same` respectively.
 
 * `one_of()` now always coerces its input to a character, or fails. Similary,
   `vars_select()` now supports unquoting S3 vectors.
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,7 +43,9 @@
   only the last one is taken into account (#52). The warning inherits
   from `tidyselect_warning_duplicate_renaming`.
 
-* `vars_select()` now ignores existing duplicate variables (#94).
+* `vars_select()` now ignores existing duplicate variables (#94). It
+  fails if user attempts to rename to an existing variable, or if
+  different variables are renamed to the same name.
 
 * `one_of()` now always coerces its input to a character, or fails. Similary,
   `vars_select()` now supports unquoting S3 vectors.
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,6 +43,10 @@
   only the last one is taken into account (#52). The warning inherits
   from `tidyselect_warning_duplicate_renaming`.
 
+* `vars_select()` now supports duplicate variables (#94). They are
+  ignored if not part of the selection. Otherwise, they are repaired
+  to make them unique.
+
 * `one_of()` now always coerces its input to a character, or fails. Similary,
   `vars_select()` now supports unquoting S3 vectors.
 
--- a/R/vars-select.R
+++ b/R/vars-select.R
@@ -269,34 +269,12 @@ inds_combine <- function(vars, inds) {
   }
 
   names(incl) <- names2(incl)
-  unnamed <- names(incl) == ""
+  unrenamed <- names(incl) == ""
+  unrenamed_vars <- vars[incl[unrenamed]]
 
-  renamers <- names(incl)[!unnamed]
+  inds_check(inds, vars, incl, dups, unrenamed, unrenamed_vars)
 
-  # Below we check that variables are renamed to a unique name. But we
-  # also want to allow renaming existing duplicates, which we remove
-  # from the checking set here.
-  if (any(dups)) {
-    ok <- names(dups)[dups]
-    renamers <- renamers[!renamers %in% ok]
-  }
-
-  if (vctrs::vec_duplicate_any(renamers)) {
-    abort(
-      "Can't rename different columns to the same column name.",
-      "tidyselect_error_rename_to_same"
-    )
-  }
-
-  unnamed_vars <- vars[incl[unnamed]]
-  if (any(vctrs::vec_in(renamers, unnamed_vars))) {
-    abort(
-      "Can't rename column to an existing column name.",
-      "tidyselect_error_rename_to_existing"
-    )
-  }
-
-  names(incl)[unnamed] <- unnamed_vars
+  names(incl)[unrenamed] <- unrenamed_vars
   incl
 }
 
@@ -347,6 +325,51 @@ ind_check <- function(x) {
 
   if (any(positive != positive[[1]])) {
     abort("Each argument must yield either positive or negative integers.")
+  }
+}
+
+inds_check <- function(x, vars, incl, dups, unrenamed, unrenamed_vars) {
+  renamers <- names(incl)[!unrenamed]
+
+  # Below we check that variables are renamed to a unique name. But we
+  # also want to allow renaming existing duplicates, which we remove
+  # from the checking set here.
+  if (any(dups)) {
+    ok <- names(dups)[dups]
+    renamers <- renamers[!renamers %in% ok]
+  }
+
+  if (vctrs::vec_duplicate_any(renamers)) {
+    dups <- vctrs::vec_duplicate_detect(renamers)
+    dups <- vctrs::vec_unique(renamers[dups])
+
+    probs <- map_chr(dups, function(dup) {
+      cols <- vars[incl[names(incl) == dup]]
+      cols <- glue::backtick(cols)
+      cols <- glue::glue_collapse(cols, sep = ", ", last = " and ")
+      glue::glue("* Columns {cols} are being renamed to `{dup}`.")
+    })
+    msg <- paste_line(
+      "Can't rename different columns to the same column name.",
+      !!!probs
+    )
+
+    abort(msg, "tidyselect_error_rename_to_same")
+  }
+
+  if (any(vctrs::vec_in(renamers, unrenamed_vars))) {
+    dups <- unrenamed_vars[match(renamers, unrenamed_vars, 0L)]
+
+    probs <- map_chr(dups, function(dup) {
+      col <- vars[[incl[[dup]]]]
+      glue::glue("* Column `{col}` is being renamed to existing column `{dup}`.")
+    })
+    msg <- paste_line(
+      "Can't rename column to an existing column name.",
+      !!!probs
+    )
+
+    abort(msg, "tidyselect_error_rename_to_existing")
   }
 }
 
--- a/R/vars-select.R
+++ b/R/vars-select.R
@@ -226,7 +226,17 @@ inds_combine <- function(vars, inds) {
   walk(inds, ind_check)
   first_negative <- length(inds) && length(inds[[1]]) && inds[[1]][[1]] < 0
 
-  inds <- vctrs::vec_c(!!!inds, .ptype = integer(), .name_spec = "{outer}{inner}")
+  # Don't suffix existing duplicate with a sequential suffix
+  dups <- purrr::map_lgl(inds, is_data_dups)
+  spec <- function(outer, inner) {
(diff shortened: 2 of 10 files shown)


lionel-:
The last commit does introduce a breaking change.

Before:

```r
tibble(a = 1, b = 2) %>% select(foo = everything())
#> # A tibble: 1 x 2
#>    foo1  foo2
#>   <dbl> <dbl>
#> 1     1     2
```

After:

```r
tibble(a = 1, b = 2) %>% select(foo = everything())
#> New names:
#> * foo -> foo...1
#> * foo -> foo...2
#> # A tibble: 1 x 2
#>   foo...1 foo...2
#>     <dbl>   <dbl>
#> 1       1       2
```

This way we get the following consistent behaviour:

```r
vars_select(c("a", "b", "a"), b = a, a = b)
#> New names:
#> * b -> b...1
#> * b -> b...2
#> b...1 b...2     a
#>   "a"   "a"   "b"
```

Not sure how disruptive this change is going to be.

lionel-:
If we are not willing to make the breaking change in the last commit, then I think it should be an error to select duplicate variables. The PR will still be an improvement because it makes it possible to repair a data frame with redundant variable.

And then for consistency, renaming to an existing variable (cf PR #116) should be an error rather than a name repair.

jennybc:
I am nodding along with everything you say and show above. I can't give a resounding endorsement, just because I feel pretty out of touch with all the possible implications of this. That is, my intuition is very much based on usage of downstream packages that call into tidyselect.

NEWS.md
```diff
@@ -1,6 +1,10 @@
 
 # tidyselect 0.2.5.9000
 
+* `vars_select()` now supports duplicate variables (#94). They are
+  ignored if not part of the selection. Otherwise, they are repaired
+  to make them unique.
+
```
jennybc: The thing that feels odd is I perceive `vars_select()` as being "read only", i.e. changing the names that it's operating on is out of scope.

lionel-: You can change the names with `data %>% dplyr::select(foo = bar)`. The main difference between `select()` and `rename()` is that the former is a shrinking verb while the latter maintains the shape.

lionel-:
@jennybc The high level view of these changes is that

1. It becomes possible to remove duplicate variables with `select()` (which is a way to repair a data frame manually). 

2. If duplicate variables are selected, their names are repaired with a message.

The second behaviour is consistent with the `rename()` changes we have discussed in the other PR, but it might cause issues with this existing pattern:

```r
tibble(a = 1, b = 2) %>% select(foo = everything())

# Before

#> # A tibble: 1 x 2
#>    foo1  foo2
#>   <dbl> <dbl>
#> 1     1     2


# After

#> New names:
#> * foo -> foo...1
#> * foo -> foo...2
#> # A tibble: 1 x 2
#>   foo...1 foo...2
#>     <dbl>   <dbl>
#> 1       1       2
```

jennybc:
Again, with the caveat that I'm focused on how this *feels* in dplyr, it feels surprising that `select()` can do something other than requested -- i.e. it can decide to repair names -- because it found duplicate names. Elsewhere, the default stance is that -- unless expressly authorized -- duplicate names is an error, i.e. `.repair_names = "check_unique"`. If you want repair, then you need to authorize it via `.repair_names = "unique"`.

lionel-:
I see. More questions:

* Then does this mean that renaming to an existing name should be an error rather than a repair?

* And then, should `select()` be used to remove duplicates, and fail with an error if some duplicates are left in the selection, rather than silently leaving them?

* Regarding the _generated_ names, i.e. `foo = everything()`, do you agree with the principle that name combining is done with `...` separator? I guess this suggests `select()` should gain a `name_spec` argument, like `vec_c()` has.

lionel-:
In fact the behaviour shouldn't be the same because we have two different situations:

1. For renaming: It is an error to create a _new_ duplicate name.

2. For selecting: We try to ignore _existing_ duplicates as much as possible.

So creating a new duplicate variable in `select()` via renaming should be an error, but selecting duplicate variables shouldn't.

jennybc:
Now you've basically convinced me `select()` should be operating as if `.name_repair = "minimal"`.

That is, we assume you know what you're doing and, if you have duplicates, it's because you want them or they involve variables you don't care/think about (and you're happy to continue in that manner).

I'm starting to think of it like this:

  * We have a higher level of rigor and requirement for explicitness at tibble formation time. So, in functions like `tibble()`, `as_tibble()`, `read_excel()`, etc.
  * But there's room and good reasons to not be so picky once we are doing sequential manipulations on an existing tibble.

Regardless, I think we should force ourselves to describe what's going to happen in terms of an existing level of name repair.

> So creating a new duplicate variable in select() via renaming should be an error, but selecting duplicate variables shouldn't. WDYT?

This sounds good to me but I worry that it's too special-case-y.

Would `select()` gain a `.name_repair` argument? That seems like a really important fact to know or decide.

jennybc:
As for ways people use `select()`, I see:

  * Selecting columns (the case where we don't select everything)
  * Reordering columns (the case where we use `everything()`)
  * Renaming columns

And, of course, a single `select()` call can express the intent to do more than one of these.

I think we should evaluate any proposal against these behaviours.

lionel-:
> That is, we assume you know what you're doing and, if you have duplicates, it's because you want them or they involve variables you don't care/think about (and you're happy to continue in that manner).

Good summary.


> We have a higher level of rigor and requirement for explicitness at tibble formation time

Or more generally, at column-creation time. In tidyselect column creation occurs with renaming, both in `rename()` and in `select()`.


> This sounds good to me but I worry that it's too special-case-y.

I see it as the consequence of the two rules above: minimal for normal usage, strict for column creation.


> Would select() gain a .name_repair argument? That seems like a really important fact to know or decide.

I would rather not open this door, it seems like this would lead us adding that argument in many places. I think the UI will be more streamlined if we're being laxist or strict when it matters, and offer a distinct verb for repairing the tibble. Which makes me wonder, do we have one?

jennybc:
> offer a distinct verb for repairing the tibble. Which makes me wonder, do we have one?

For name-repairing? I guess you just do `as_tibble()` again.

``` r
library(tidyverse)

df <- tibble(x = 1, y = 2)

rename(df, x = y)
#> # A tibble: 1 x 2
#>       x     x
#>   <dbl> <dbl>
#> 1     1     2

rename(df, x = y) %>% as_tibble()
#> Column name `x` must not be duplicated.
#> Use .name_repair to specify repair.

rename(df, x = y) %>% as_tibble(.name_repair = "unique")
#> New names:
#> * x -> x...1
#> * x -> x...2
#> # A tibble: 1 x 2
#>   x...1 x...2
#>   <dbl> <dbl>
#> 1     1     2
```

<sup>Created on 2019-09-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

lionel-:
I think we now have solid semantics.

* We throw a `tidyselect_error_rename_to_same` error when multiple variables are renamed to the same name:

    ```r
    vars_select(letters, foo = a, foo = b)
    #> Error: Must use different names when renaming columns.
    #> * Columns `a` and `b` are being renamed to `foo`.

    vars_select(letters, foo = a, bar = b, foo = c, ok = d, bar = e)
    #> Error: Must use different names when renaming columns.
    #> * Columns `a` and `c` are being renamed to `foo`
    #> * Columns `b` and `e` are being renamed to `bar`
    ```


* We throw a `tidyselect_error_rename_to_existing` error when a variable is renamed to an existing column name:

    ```r
    vars_select(letters, a, a = b)
    #> Error: Must use new name when renaming columns.
    #> * Column `b` is being renamed to existing column `a`.

    vars_select(letters, a = b, ok = c, d = e, everything())
    #> Error: Must use new name when renaming columns.
    #> * Column `b` is being renamed to existing column `a`.
    #> * Column `e` is being renamed to existing column `d`.
    ```

* However existing duplicate names are fully supported:

    ```r
    vars <- c("a", "b", "a")

    vars_select(vars, b, starts_with("a"))
    #>   b   a   a
    #> "b" "a" "a"
    ```

*   They can also be renamed:

    ```r
    vars_select(vars, b = a, a = b)
    #>   b   b   a
    #> "a" "a" "b"
    ```

* And we can still rename with sequence suffix:

    ```r
    vars_select(letters, foo = c(a, b, c))
    #> foo1 foo2 foo3
    #>  "a"  "b"  "c"
    ```

lionel-:
>  I guess you just do as_tibble() again.

That makes sense thanks!

tests/testthat/test-vars-select.R
```diff
@@ -155,3 +155,59 @@ test_that("vars_select() supports S3 vectors (#109)", {
...
+  expect_error(vars_select(letters[1:2], a, a = b), class = "tidyselect_error_rename_to_existing")
+})
+
+test_that("vars_select() fails when renaming to same name", {
+  expect_error(vars_select(letters[1:3], a = b, a = c), class = "tidyselect_error_rename_to_same")
+  expect_error(vars_select(letters[1:2], A = a, A = b), class = "tidyselect_error_rename_to_same")
+})
+
+test_that("vars_select() fails informatively", {
+  expect_known_output(file = test_path("outputs", "vars-select-renaming-to-same.txt"), {
```
hadley: You should be able to use `verify_output()` here so that you don't need your own `try()`.

R/vars-select.R
```diff
@@ -311,6 +333,51 @@ ind_check <- function(x) {
...
+    dups <- vctrs::vec_unique(renamers[dups])
+
+    probs <- map_chr(dups, function(dup) {
+      cols <- vars[incl[names(incl) == dup]]
+      cols <- glue::backtick(cols)
+      cols <- glue::glue_collapse(cols, sep = ", ", last = " and ")
+      glue::glue("* Columns {cols} are being renamed to `{dup}`.")
+    })
+    msg <- paste_line(
+      "Must use different names when renaming columns.",
```
jennybc:
```suggestion
      "Must use unique names when renaming columns.",
```

I think it's best to re-use existing name repair vocabulary and levels whenever we can.

lionel-: good point thanks!

lionel-:
> I just feel like this behaviour is going to be pretty hard to document or predict, i.e. it would cause someone to carefully re-read docs, which is usually a bad sign.

I think it is easy to document: we tolerate existing duplicates but you can't create new ones. We probably don't need to go into details because these are quite obscure corner cases.

I agree the behaviour would be clearer if we used check-unique everywhere, and that was my first inclination. If we're going to tolerate duplicates, the behaviour will necessarily be a little obscure and will depend on each verb. I think the most important thing is to follow a cardinal principle: we can't create new duplicates. It seems reasonable to consider that renaming duplicates doesn't create new ones.

> Would it help to mentally or literally decompose the above into a renaming step and a reordering step?

This sounds right, ordering of selections is determined by the ordering of selectors. Providing a LHS causes renaming.

> Why not use the ...j method of numerically disambiguating here? The reasoning is to reuse a convention we're already using elsewhere. One of the main motivations for doing name repair was to eliminate package-to-package idiosyncrasies about how names were made unique.

I asked above whether it'd be worth it to break backward compatibility to follow the new convention. I imagine this would break quite a few user scripts. Hadley later said on Slack it's probably not worth it. WDYT?

jennybc:
OK a colloquial explanation of the name repair situation is "you aren't allowed to make things worse". I still think it's weird that we can rename existing sets of duplicates, but this is also a very edge case and I can let it go.

Ditto about the numeric suffix style. It seems counter to the consistency goals to continue to implement `foo1, foo2, ...` instead of `foo...1, foo...2, ...`. But I can let it go.

lionel-:
> OK a colloquial explanation of the name repair situation is "you aren't allowed to make things worse".

Nice way of putting it.

Thanks for the helpful reviews Jenny!

PR #124, opened 2019-10-15
Use vctrs to validate indices and positions

lionel-:
Requires r-lib/vctrs#616.
Closes #119.

Following is a selection of vctrs-powered errors. They seem to be consistent and informative. All these errors inherit from `tidyselect_error_index`.

1. `vars_select()`

```r
"Bare names"

vars_select(letters, foo)
#> Error: Must select existing columns.
#> x Can't subset element with unknown name `foo`.

vars_select(letters, -foo)
#> Error: Must select existing columns.
#> x Can't subset element with unknown name `foo`.


"Names"

vars_select(letters, "foo")
#> Error: Must select existing columns.
#> x Can't subset element with unknown name `foo`.

vars_select(letters, a:"foo")
#> Error: Must select existing columns.
#> x Can't subset element with unknown name `foo`.


"Positions"

vars_select(letters, 30, 50, 100)
#> Error: Must select existing columns.
#> x Can't subset positions 30, 50 and 100.
#> i There are only 26 elements.

vars_select(letters, -100)
#> Error: Must select existing columns.
#> x Can't subset position 100.
#> i There are only 26 elements.

vars_select(letters, 2.5)
#> Error: Must select with column names or positions.
#> x Lossy cast from `i` <double> to <integer>.
```

2. `vars_rename()`

```r
vars_rename(letters, A = TRUE)
#> Error: Must select with column names or positions.
#> x `i` has the wrong type `logical`.
#> i Positions and names must be integer or character.

vars_rename(letters, A = 1.5)
#> Error: Must select with column names or positions.
#> x Lossy cast from `i` <double> to <integer>.
```

3. `vars_pull()`

```r
vars_pull(letters, letters)
#> Error: Must select with column names or positions.
#> x `var` has the wrong size 26.
#> i Positions and names must be size 1.

vars_pull(letters, NA)
#> Error: Must select with column names or positions.
#> x `var` can't be `NA`.
#> i Positions and names can't be missing.

vars_pull(letters, "foo")
#> Error: Must select existing columns.
#> x Can't subset element with unknown name `foo`.

vars_pull(letters, -100)
#> Error: Must select existing columns.
#> x Can't subset position 100.
#> i There are only 26 elements.

vars_pull(letters, -Inf)
#> Error: Must select with column names or positions.
#> x Lossy cast from `var` <double> to <integer>.
```

`vars_rename()` and `vars_pull()` currently still throw "object not found" errors rather than OOB errors.

PR #125, opened 2019-10-17
Add `all_of()` and `any_of()`

lionel-:
Depends on #124, r-lib/vctrs#615 and r-lib/vctrs#616.


R/select-helpers.R
```diff
@@ -107,27 +120,34 @@ num_range <- function(prefix, range, width = NULL, vars = peek_vars()) {
...
 
-#' @export
 #' @rdname select_helpers
-#' @param ... One or more character vectors.
-one_of <- function(..., .vars = peek_vars()) {
-  keep <- compact(list(...))
+#' @param x A character vector.
+#' @inheritParams ellipsis::dots_empty
+#' @export
+all_of <- function(x, ..., vars = peek_vars()) {
```
hadley: Why do we allow dots here? Or conversely, why don't we use them?

lionel-: They are not allowed because of the ellipsis check. They are forcing `vars()` to be passed by name.

lionel-: It doesn't seem useful to take data by dots. This is a way of being prudent and conservative with the function signature.

hadley: :+1: can you make that clear in the news? (ie we’re being more conservative so the equivalent of `one_of(“a”, “b”)` is `all_of(c(“a”, “b”))`)

lionel-: Done in https://github.com/r-lib/tidyselect/commit/372eaf29f4244050d5e6a3b8308dddbccaa95247

PR #131, opened 2019-10-24
Implement `select_pos()`

### LINKED ISSUE #86: Support functions in selections
Once we have https://github.com/r-lib/rlang/issues/726, we can extend the selection DSL with predicate functions. Those would expand to a vector of column positions.

```r
data %>% select(is.numeric)

data %>% select(-c(is.numeric, one_of(nms)))
```

This would make the `_if` scoped variants in dplyr largely irrelevant. Also the `_all` variants are just `_at(vars(everything())`, so if we came up with a more concise syntax we wouldn't need those either. The successors of scoped variants could be just one function.

cc @romainfrancois @hadley

lionel-:
Part of #89 and closes #86.

* Takes a vector or data frame and returns a vector of selected positions. Because it takes the whole data as input, it supports predicate selections. Because it returns a vector of positions rather than names, it has better support for duplicated names.

* Unexported `vec_select()` serves as an implementation example. While `vec_select()` has vctrs semantics (but refuses to work with data frames), `select_pos()` has `[` semantics and works with data frames. This should make it easier for packages like tidyr to interface with it.

* Selections are taken via `...` for now, because `c(...)` isn't consistent with `...` yet.

* Will be documented and exported later on when `c(...)` has been fixed. We will switch to a single named parameter to encourage tidyr-like interfaces, which should be more common than dplyr-like ones.

### DIFF
--- a/R/select.R
+++ b/R/select.R
@@ -1,12 +1,22 @@
 
-select_pos <- function(.x, ..., .strict = TRUE) {
+select_pos <- function(.x,
+                       ...,
+                       .include = NULL,
+                       .exclude = NULL,
+                       .strict = TRUE) {
   vars <- names(.x)
   if (is_null(vars)) {
     abort("Can't select within an unnamed vector.")
   }
 
   scoped_vars(vars)
-  select_impl(.x, ..., .strict = .strict)
+  select_impl(
+    .x,
+    ...,
+    .include = .include,
+    .exclude = .exclude,
+    .strict = .strict
+  )
 }
 
 # Example of implementation, mainly used for unit tests
@@ -20,23 +30,32 @@ vec_select <- function(.x, ..., .strict = TRUE) {
 
 select_impl <- function(.x = NULL,
                         ...,
+                        .include = NULL,
+                        .exclude = NULL,
                         .strict = TRUE) {
   dots <- enquos(...)
   vars <- peek_vars()
-
-  ind_list <- subclass_index_errors(
-    vars_select_eval(vars, dots, .strict, data = .x)
-  )
 
   # If the first selector is exclusive (negative), start with all
   # columns. We need to check for symbolic `-` here because if the
   # selection is empty, `inds_combine()` cannot detect a negative
   # indice in first position.
-  if (is_negated(quo_get_expr(dots[[1]]))) {
-    ind_list <- c(list(seq_along(vars)), ind_list)
+  if (length(dots) && is_negated(dots[[1]])) {
+    dots <- list(expr(c(everything(), !!!dots)))
   }
 
-  inds_combine(vars, ind_list)
+  if (length(.include)) {
+    dots <- list(expr(all_of(!!.include) | c(!!!dots)))
+  }
+  if (length(.exclude)) {
+    dots <- list(expr(c(!!!dots, -!!.exclude)))
+  }
+
+  inds <- subclass_index_errors(
+    vars_select_eval(vars, dots, .strict, data = .x)
+  )
+
+  inds_combine(vars, inds)
 }
 
 is_negated <- function(x) {
--- a/R/select.R
+++ b/R/select.R
@@ -22,12 +22,9 @@ select_pos <- function(.x,
 }
 
 # Example of implementation, mainly used for unit tests
-vec_select <- function(.x, ..., .strict = TRUE) {
-  if (is.data.frame(.x)) {
-    abort("`.x` can't be a data frame.")
-  }
+select <- function(.x, ..., .strict = TRUE) {
   idx <- select_pos(.x, ..., .strict = .strict)
-  set_names(vctrs::vec_slice(.x, idx), names(idx))
+  set_names(.x[idx], names(idx))
 }
 
 select_impl <- function(.x = NULL,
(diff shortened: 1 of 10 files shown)


R/vars-select-eval.R
```diff
@@ -263,18 +283,32 @@ eval_sym <- function(name, data_mask, context_mask, colon = FALSE) {
...
-    inform(glue_c(
-      "Note: Using an external vector in selections is brittle.",
-      i = "If the data contains `{name}` it will be selected instead.",
-      i = "Use `all_of({name})` instead of just `{name}` to silence this message."
-    ))
+    if (!is_function(value)) {
+      inform(glue_c(
+        "Note: Using an external vector in selections is brittle.",
+        i = "If the data contains `{name}` it will be selected instead.",
+        i = "Use `all_of({name})` instead of just `{name}` to silence this message."
```
hadley: Drive by:

```suggestion
        i = "Use `all_of({name})` instead of `{name}` to silence this message."
```

tests/testthat/outputs/vars-select-bool-symbols.txt
```diff
@@ -1,10 +1,6 @@
 > vars_select(letters, starts_with("a") & z)
-Error: Can't use boolean operators with bare variables.
-x `z` is a bare variable.
-i Do you need `all_of(z)`?
+Error in eval_tidy(expr, context_mask): object 'z' not found
```
hadley: Is this correct?

R/select.R
```diff
@@ -0,0 +1,74 @@
...
+  )
+}
+
+# Example of implementation, mainly used for unit tests
+vec_select <- function(.x, ..., .strict = TRUE) {
+  if (is.data.frame(.x)) {
+    abort("`.x` can't be a data frame.")
+  }
+  idx <- select_pos(.x, ..., .strict = .strict)
+  set_names(vctrs::vec_slice(.x, idx), names(idx))
```
hadley: Maybe just use `[` here? Given that it's only an example and that would make it work with data frames, which would be nice for testing?

R/select.R
```diff
@@ -0,0 +1,74 @@
...
+}
+
+select_impl <- function(.x = NULL,
+                        ...,
+                        .include = NULL,
+                        .exclude = NULL,
+                        .strict = TRUE) {
+  dots <- enquos(...)
```
hadley: Would it be worth making this `vars <- c(!!!enquos(...))`? It would simplify the `.include` and `.exclude` implementation and would reduce possible inconsistencies.

lionel-: Yup this is where we'll end up once `c(...)` is consistent with current `...` behaviour. I need to fix input joining, and duplicate name checking.

tests/testthat/outputs/vars-select-bool-symbols.txt
lionel-: Hmm I'm not sure anymore. Also I now realise the previous advice was not correct either (the `all_of(z)` wouldn't have worked, since `z` is a data variable here, not a context variable).

Here are the constraints:

- We disallow data lookup because users might be tempted to use `cyl & disp` where `cyl | disp` would be correct.
- We disallow context lookup of external vectors: user has to use `all_of()` in that case.
- We allow context lookup of predicate functions.

So what should be the error?

1. If we found an object that is not a function, we could throw a bad type error if the type is not right, or recommend `all_of()` if the type is adequate.

2. If we didn't find any object, the default "object not found" error seems ok.

Does that make sense? Currently we have error 2 but we're missing error 1.

R/select.R
hadley: Ok, cool

tests/testthat/outputs/vars-select-bool-symbols.txt
hadley: Hmmmm, I think it might be ok to allow bare variables here because (e.g.) `a | b | c` reads quite nicely and presumably `"a" | "b" | "c"` is ok? OTOH that introduces ambiguity if you have columns called `is.numeric()` so I think you're right.

I think it probably needs to be a specific error (as before) but with a hint that works — I guess the big question is what that should be. Maybe `starts_with("a") & c(z)`? But then that implies that `c(is.numeric)` would look for variable `is.numeric` which would be inconsistent with a bare `is.numeric`.

Or do we just say that variable names beat predicate names, and that's ok because it's sufficiently unusual to call a variable (e.g.) `is.numeric`?

lionel-: Perhaps a good principle to follow is that `select(a, b)` and `select(a | b)` should have exactly the same properties, otherwise it will be awkward to explain.

Also I realised that allowing full data + function scope in `a & b` can be useful too, for example `foo & is.numeric` means "select variable only if called foo and only if numeric".

The main question is what kind of error do we want when `a` is not a data frame variable nor a function. Should this be an object not found error, or an out-of-bounds error? One advantage of requiring wrapping in something like `where()` is that we solve the scoping ambiguity, and then having the OOB error is unambiguous as well.

lionel-:
We now:

- Allow referring to columns with symbols in boolean operands.
- Throw OOB errors if a symbol operand doesn't exist.
- Throw an informative error when the intersection of two columns is taken.
- Throw a bad index type error if a function is supplied to `all_of()`.

PR #134, opened 2019-11-07
Consolidate dots syntax

lionel-:
- Unary `-` inside dots is syntax for set difference once again. This is also the case inside `c()`.

- Set operations are redefined to include names. `c(foo = a)` and `c(bar = a)` are different elements. Unnamed elements match any named ones, which makes it possible to rename one element from a set of columns with `starts_with("Sepal"), foo = Sepal.Width`.

- The old `inds_combine()` function is no longer needed and has been removed.

- We no longer try to preserve existing duplicates because this makes things too complex. It is still possible to repair a data frame by selecting unique elements, but duplicates inside a selection now cause an error.

- I have started to write up a technical description of the tidyselect syntax. It is currently a doc topic at the bottom of select.R but I'll move it to a vignette when I finish it up.


NEWS.md
```diff
@@ -1,8 +1,12 @@
 
 # tidyselect (development)
 
-* Improved detection of negated arguments in first position, including
-  inside `c()`.
+* Unary `-` inside nested `c()` is now consistently syntax for set
```
hadley: What do you mean by consistently?

R/select-sets.R
```diff
@@ -0,0 +1,77 @@
...
+}
+sel_intersect <- function(x, y) {
+  if (is_null(names(x)) && is_null(names(y))) {
+    set_intersect(x, y)
+  } else {
+    sel_operation(x, y, set_intersect)
+  }
+}
+sel_unique <- function(x) {
+  x <- tibble::as_tibble(list(value = x, names = names2(x)))
```
hadley: Why not `vctrs::new_data_frame()`?

NEWS.md
hadley: Presumably this also means that `c(-y)` works, where it didn't before?

tests/testthat/outputs/c-rename-duplicates.txt
```diff
@@ -0,0 +1,4 @@
+> select_pos(df, c(foo = a))
```
hadley: Maybe include `names(df)` here too?

tests/testthat/outputs/vars-select-oob-errors.txt
```diff
@@ -19,7 +19,7 @@ x Can't subset element with unknown name `foo`.
 > # Positions
 > vars_select(letters, 30, 50, 100)
 Error: Must select existing columns.
-x Can't subset positions 30, 50 and 100.
+x Can't subset position 100.
```
hadley: Was this a deliberate change?

tests/testthat/test-vars-rename.R
```diff
@@ -26,8 +25,7 @@ test_that("when .strict = FALSE, vars_rename always succeeds", {
 
   expect_error(
     vars_rename(c("a", "b"), d = "e", f = "g", .strict = TRUE),
-    "Unknown columns `e` and `g`",
-    fixed = TRUE
+    class = "tidyselect_error_index_oob_names"
```
hadley: Do you think it's worth testing for that the error contains e and g?

tests/testthat/outputs/c-rename-duplicates.txt
lionel-: Good idea.

tests/testthat/outputs/vars-select-oob-errors.txt
lionel-: Yes but I found a way to get the original message back. Early validation is only needed for the RHS of set-difference (as OOB values might be filtered out). We now validate lazily for other operations to get a more complete error report.

tests/testthat/test-vars-rename.R
lionel-: It now only contains `e`. We'll get a more informative error when we merge rename and select implementations. Except for character inputs since those need to be validated early.

I have added a `verify_output()` to track these messages.

PR #136, opened 2019-11-08
Sanitise base errors

lionel-:
This prevents the function calls from leaking into base errors during evaluation.

Before:

```r
#> Error in eval_tidy(quo, mask): object 'e' not found
```

After:

```r
#> Error: object 'e' not found
```

We should probably use this approach in all data masking APIs, such as in dplyr. cc @batpigandme and @romainfrancois


lionel-:
As a bonus all base errors get an rlang backtrace.

PR #139, opened 2019-11-13
Export and document `eval_select()` and `eval_rename()`

### LINKED ISSUE #87: Document how to use tidyselect in a new package
`use_tidy_select()`?

### LINKED ISSUE #89: Tidyselect functions should take a prototype as input

### LINKED ISSUE #93: rename_all fails on data.frame with dirty names
`rename_all` requires distinct, non-missing variable names in the input data.frame.  This is something of a circular problem for me: I want to use `rename_all` in `janitor::clean_names()` to address https://github.com/sfirke/janitor/issues/260 where grouping indices of a grouped_df are not updated along with the column names.

But the problematic variable names that often go into `clean_names()` are causing `rename_all` to fail.  For instance, with duplicated names, or missing names:

```r
# repeat names
test_df <- as.data.frame(matrix(ncol = 2))
names(test_df) <- c("repeat", "repeat")
test_df %>% rename_all(make.names)
#> `data` must be uniquely named but has duplicate columns

# missing names
test_df <- as.data.frame(matrix(ncol = 2))
names(test_df) <- c("hi", "")
test_df %>% rename_all(make.names)
#> Error in is_symbol(expr): argument "expr" is missing, with no default
```
(I'm using `make.names` to show it's not a `make_clean_names()` problem, but my actual use case would be calling `make_clean_names()`)
 
I welcome advice of all kinds - if there's a way I can get `clean_names()` to update the group names of a grouped_df without using `rename_all` that's great too 😃

lionel-:
Closes #87
Closes #89
Closes #93

- Export and document eval functions.
- Add a section about renaming in the syntax vignette.
- New Get started vignette, rendered at http://rpubs.com/lionel-/tidyselect-get-started


vignettes/syntax.Rmd
```diff
@@ -27,9 +27,15 @@ This is a technical description of the tidyselect syntax.
 library(tidyselect)
 library(magrittr)
 
-# FIXME: `select_pos()` should be exported as `eval_select()`
+# FIXME: `eval_select()` and `eval_rename()` should be exported
```
hadley: Can remove this now?

vignettes/syntax.Rmd
```diff
@@ -462,6 +468,67 @@ mtcars %>% select_pos(mpg * wt)
...
+
+However, the name conflict can be solved by renaming the existing
+variable to another name:
+
+```{r}
+mtcars %>% select_pos(foo = cyl, cyl = mpg)
+
+mtcars %>% rename_pos(foo = cyl, cyl = mpg)
+```
+
```
hadley: Is it worth mentioning something about inputs with existing duplicates?

vignettes/tidyselect.Rmd
```diff
@@ -0,0 +1,310 @@
...
+  tibble.print_min = 4,
+  tibble.print_max = 4
+)
+```
+
+tidyselect implements a specialised sublanguage of R for selecting
+variables from data frames and other data structures. A technical
+description of the DSL is available in the [syntax
+vignette](https://tidyselect.r-lib.org/articles/syntax.html). In this
+vignette, we describe how to include tidyselect in your own packages.
```
hadley: If you just want to know how to use tidyselect syntax in dplyr or tidyr, please read xxxxxxxx instead.

vignettes/tidyselect.Rmd
```diff
@@ -0,0 +1,310 @@
...
+
+    ```{r, eval = FALSE}
+    mtcars %>% pivot_longer(c(mpg, cyl))
+    mtcars %>% pivot_longer(mpg | cyl)
+    ```
+
+Our general recommendation is to take dots when the main purpose of
+the function is to create a new data structure based on a selection.
+When the selection is accessory to the main purpose of the function,
+take it as a named argument. For more advice about this, see the
```
hadley: If in doubt, we recommend using named arguments, because it's easier to go from named arg -> dots than dots -> named arg.

vignettes/tidyselect.Rmd
```diff
@@ -0,0 +1,310 @@
...
+[Making data with `...`](https://design.tidyverse.org/dots-data.html)
+section of the tidyverse design book.
+
+
+### Do you need tidyselect?
+
+The tools described in this vignette are rather low level. Depending
+on your use case, it may be easier to wrap around `dplyr::select()`.
```
hadley:
```suggestion
on your use case, it may be easier to wrap `dplyr::select()`.
```

vignettes/tidyselect.Rmd
```diff
@@ -0,0 +1,310 @@
...
+You'll get a data frame containing the columns selected by your user,
+which you can then handle in various ways.
+
+The following examples illustrate how you could write a function that
+takes a selection of data and returns the corresponding data frame
+with capitalised names:
+
+```{r, eval = FALSE}
+# Passing dots
```
hadley: These should use precisely the same terms as above

vignettes/tidyselect.Rmd
```diff
@@ -0,0 +1,310 @@
...
+inside the input data. If you don't need the selected positions and
+can afford the dependency, you may consider wrapping around dplyr
+instead.
+
+
+## The selection evaluators
+
+tidyselect is implemented with non-standard evaluation (NSE). This
+unique feature of the R language refers to the ability of functions to
+__defuse__ some or all of their arguments, and resume evaluation later
```
hadley: Maybe use "delay execution" as well?

lionel-: I thought we were going for "defuse", from the booby trap analogy?

vignettes/tidyselect.Rmd
lionel-: I'll add a TODO because we don't currently have a user-friendly description of the tidyselect syntax. Maybe we should make a cheatsheet instead of a vignette? Or both?

If we make a vignette, should it live in tidyselect or dplyr?

vignettes/syntax.Rmd
lionel-: Added a section about duplicate columns in data frames.

vignettes/tidyselect.Rmd
lionel-: hmm or we could link to this in the meantime?
https://r4ds.had.co.nz/transform.html#select

vignettes/tidyselect.Rmd
hadley: Sorry, I meant additionally as a parenthetical since not everyone will know about the new terminology

lionel-: Now mentions "delay execution", and has a footnote comparing to "quoting".

PR #140, opened 2019-11-13
Soft-deprecate `vars_select()` and `vars_rename()`

lionel-:
cc @hadley 

We currently keep `vars_pull()` as is, because it's not clear what are the ideal semantics, and it's not a very important function.


hadley:
Do we want to retire instead? Isn't soft deprecating going to be a bit aggressive for dpyr?

lionel-:
I'll mark them as questioning so we can reassess whether to retire or deprecate in the next minor version.

PR #143, opened 2019-11-14
Document creation of custom selection helpers

lionel-:
- Document `peek_vars()` on its own so the reader is not distracted by `poke_vars()` etc.

- Add Get started section about creating custom helpers.


R/vars.R
```diff
@@ -1,3 +1,44 @@
+#' Peek variables in the selection context
```
hadley:
```suggestion
#' Peek at variables in the selection context
```

PR #145, opened 2019-11-15
Add faqs for frequent messages and notes

lionel-:
- Adds two documentation topics `faq-selection-context` and `faq-external-vector` to document two frequent messages encountered (or expected to be) by users.

- The faq topics are indexed under a common heading in the reference. Because the titles of the topics are the error messages, they are included in the index.

- The condition messages refer to the faq pages on the pkgdown site.

To work around a roxygen limitation, I've manually altered the rendered .Rd so the knitted Rmd is included in `@description` instead of `@details`. Expanding in details causes a distracting Details section in the faq pages. Filed an issue at https://github.com/r-lib/roxygen2/issues/970.


lionel-:
This makes me realise that instead of "brittle", we should say "ambiguous" here:

```r
#> Note: Using an external vector in selections is brittle.
```

R/faq.R
```diff
@@ -0,0 +1,17 @@
+#' Error: Must be used within a *selecting* function
+#'
+#' @description
+#'
+#' @includeRmd man/faq/selection-context.Rmd
+#'
```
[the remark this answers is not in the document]
lionel-: I tried this but reverted because the automatic cross-linking didn't seem very useful. It seemed to imply that relevant information would be found in the links:

```
\seealso{
Other faq: 
\code{\link{faq-selection-context}}
}
\concept{faq}
```

R/faq.R
```diff
@@ -0,0 +1,17 @@
+#' Error: Must be used within a *selecting* function
```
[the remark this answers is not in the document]
lionel-: I think it's useful for error messages to mention "Error:". It gives context in the FAQ section of the reference, as well as when visiting the page.

Which of these looks best?

```
FAQ - Error: Must be used within a *selecting* function

FAQ: Must be used within a *selecting* function (error message)

Error: Must be used within a *selecting* function (FAQ)
```

man/faq/external-vector.Rmd
```diff
@@ -0,0 +1,68 @@
+
+```{r, child = "setup.Rmd", include = FALSE}
+```
+
+## Ambiguity between columns and external variables
+
+In selecting functions like `dplyr::select()` or `tidyr::pivot_longer()`,
```
batpigandme:
```suggestion
With selecting functions like `dplyr::select()` or `tidyr::pivot_longer()`,
```

batpigandme: ^^ this is English being weird in that this _could_ mean "In (you, the reader) selecting (as in choosing to use) functions like..."

R/faq.R
```diff
@@ -0,0 +1,15 @@
...
+#'
+#' @includeRmd man/faq/selection-context.Rmd
+#'
+#' @name faq-selection-context
+#' @encoding UTF-8
+NULL
+
+#' Note: Using an external vector in selections is brittle
```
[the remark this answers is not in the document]
batpigandme: Disregard, I see the `tidyselect_verbose` option in the FAQ.

man/faq/external-vector.Rmd
```diff
@@ -0,0 +1,68 @@
...
+
+```{r}
+mtcars %>% select(cyl, am, vs)
+
+mtcars %>% select(mpg:disp)
+```
+
+For historical reasons, it is also possible to refer an external
+vector of variable names. You get the correct result, but with a note
+informing you that selecting with an external variable is brittle.
```
batpigandme: Would it be worth unpacking "brittle" here? (You describe the why in the context, but maybe a parenthetical aside?)

man/faq/external-vector.Rmd
```diff
@@ -0,0 +1,68 @@
...
+```
+
+```{r}
+vars <- c("cyl", "am", "vs")
+result <- mtcars %>% select(vars)
+```
+
+This note will become a warning in the future, and then an error. We
+have decided to deprecate this particular approach to using external
+vectors because they introduce an ambiguity. Imagine that the data
```
batpigandme:
```suggestion
vectors because they introduce ambiguity. Imagine that the data
```

man/faq/selection-context.Rmd
```diff
@@ -0,0 +1,25 @@
+
+```{r, child = "setup.Rmd", include = FALSE}
+```
+
+Functions like `starts_with()`, `contains()` or `matches()` are
+__selection helpers__ that only work in a selection context.
+
+Examples of valid selection contexts are:
```
batpigandme: This makes me wonder if we should turn your talk from last year's rstudio::conf into a blog post which we could point to here (or at least the part on selection versus action verbs).

batpigandme: Also, if we're not moving away from the `*_at()` verbs, it might be helpful to add them to this list.

R/faq.R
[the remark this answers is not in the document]
lionel-: We show it once per variable per session, and only if called from the global environment.

man/faq/selection-context.Rmd
lionel-: We'll move to `across()` in the next dplyr version, but that's definitely a good example of context to add in the future.

man/faq/external-vector.Rmd
lionel-: This is now:

> You get the correct result, but with a note informing you that selecting with an external variable is ambiguous because it is not clear whether you want a data frame column or an external object.

lionel-:
Thanks for the reviews!

PR #205, opened 2020-07-17
Unname the result indices of where()

### LINKED ISSUE #204: eval_rename() interprets where() incorrectly
If I understand the spec correctly, the first example below should return `c(var1 = 1, var2 = 2)`, and the second example should be errored because the `expr` doesn't have names.

``` r
library(tidyselect)

expr <- rlang::expr(c(var = where(is.numeric)))
eval_rename(expr, list(a = 1, b = 2, c = "x"))
#> var...a var...b 
#>       1       2

expr_nonames <- rlang::expr(where(is.numeric))
eval_rename(expr_nonames, list(a = 1, b = 2, c = "x"))
#> a b 
#> 1 2
```

<sup>Created on 2020-07-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

yutannihilation:
Fix #204

### DIFF
--- a/R/eval-walk.R
+++ b/R/eval-walk.R
@@ -144,7 +144,7 @@ as_indices_sel_impl <- function(x, vars, strict, data = NULL) {
       ))
     }
     predicate <- x
-    x <- which(map_lgl(data, predicate))
+    x <- which(map_lgl(unname(data), predicate))
   }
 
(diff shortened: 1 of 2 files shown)


yutannihilation:
Closing in favor of #246

Issue #211, opened 2020-09-06
Conflict between selection of multiple columns in `last_col()` and _R_CHECK_LENGTH_1_LOGIC2_


yjunechoe:
When you have set `"_R_CHECK_LENGTH_1_LOGIC2_"` to `TRUE` (which many folks do to guard against accidentally using `&&` for non-scalars), you are no longer able to select multiple columns with `last_col()`.

Here's an example where I select the last three columns of `starwars` by passing in the vector `0:2` as the first argument to `last_col()`. It works fine normally but fails when the environment variable is set to `TRUE`

``` r
library(dplyr)

Sys.setenv("_R_CHECK_LENGTH_1_LOGIC2_" = FALSE)

starwars %>%
  select(last_col(0:2))
#> # A tibble: 87 x 3
#>    starships vehicles  films    
#>    <list>    <list>    <list>   
#>  1 <chr [2]> <chr [2]> <chr [5]>
#>  2 <chr [0]> <chr [0]> <chr [6]>
#>  3 <chr [0]> <chr [0]> <chr [7]>
#>  4 <chr [1]> <chr [0]> <chr [4]>
#>  5 <chr [0]> <chr [1]> <chr [5]>
#>  6 <chr [0]> <chr [0]> <chr [3]>
#>  7 <chr [0]> <chr [0]> <chr [3]>
#>  8 <chr [0]> <chr [0]> <chr [1]>
#>  9 <chr [1]> <chr [0]> <chr [1]>
#> 10 <chr [5]> <chr [1]> <chr [6]>
#> # ... with 77 more rows
  
Sys.setenv("_R_CHECK_LENGTH_1_LOGIC2_" = TRUE)

starwars %>%
  select(last_col(0:2))
#> Error: 'length(x) = 3 > 1' in coercion to 'logical(1)'
 ```

<sup>Created on 2020-09-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

And I can see that the source code for `last_col()` uses `&&` in the first condition, which checks that **A)** a number is passed in to the `offset` argument and **B)** this number does not exceed the total number of columns.  This line here seems to be triggering the conflict

https://github.com/r-lib/tidyselect/blob/7f3b5bb8ac45a355bbe5ace6732bda972fd40eb6/R/helpers-misc.R#L69

[...]

PR #257, opened 2022-02-03
add protection against passing vector offset=

MichaelChirico:
As used mistakenly in a test here:

https://github.com/tidymodels/broom/issues/1077


hadley:
Fixes #211

hadley:
I've made a merge conflict, and I want to take a slightly different approach, so I'll take this on. Thanks for working on it!

PR #273, opened 2022-08-10
Reorganise tests

hadley:


tests/testthat/test-vars.R
```diff
@@ -60,3 +60,13 @@ test_that("full data is in scope", {
 test_that("peek_data() fails informatively", {
   expect_error(peek_data(), "must be used within")
 })
+
+test_that("generic error message is thrown if `fn` is not supplied", {
+  local_vars(NULL)
```
hadley, line 65: I had to add `local_vars(NULL)` to get these tests to work — I assume that's because there's some unintended leakage across tests, but I spent 5 minutes looking and couldn't find it.

hadley: Oh, I can fix it by switching out the homegrown code in `peek_vars()` for `withr::defer()`. Let me know if you want me to do that (will need to move withr from suggests to imports).

lionel-: The other way we could do this is by copying the compat file in tidyselect, though I think it's better to minimise the number of packages that do that. Internal changes to the handlers structures cause incompatibilities between packages until they update the compat file.

I think it's probably fine for tidyselect to depend on withr. I think if it were up to me withr would fetch `defer()` from rlang, but Jim probably had a good reason not to take a dependency on rlang.

lionel- (approved):

PR #275, opened 2022-08-10
Detect stringr pattern in matches()

### LINKED ISSUE #238: matches() should optionally use stringr
If `match` has class `pattern` (which I should probably change to `stringr_pattern`, https://github.com/tidyverse/stringr/issues/384)

(Low priority, but will make `matches()` a bit more consistent with the general direction in which we're going)

hadley:
Fixes #238


DESCRIPTION
```diff
@@ -30,6 +30,7 @@ Suggests:
     rmarkdown,
     testthat (>= 3.1.1),
     tibble (>= 2.1.3),
+    stringr
```
lionel-, line 33: Nit: Should be alphabetically sorted.

R/helpers-pattern.R
```diff
@@ -144,7 +146,23 @@ matches <- function(match,
                     vars = NULL) {
   check_match(match)
   vars <- vars %||% peek_vars(fn = "matches")
-  flat_map_int(match, grep_vars, vars, ignore.case = ignore.case, perl = perl)
+
+  if (inherits(match, "pattern") || inherits(match, "stringr_pattern")) {
+    check_installed("stringr")
+    if (!missing(ignore.case)) {
+      cli::cli_abort("{.arg ignore.case} not supported when {.arg match} is a stringr pattern")
```
lionel-, line 153:
```suggestion
      cli::cli_abort("{.arg ignore.case} not supported when {.arg match} is a stringr pattern.")
```

R/helpers-pattern.R
```diff
@@ -144,7 +146,23 @@ matches <- function(match,
...
+    }
+    if (!missing(perl)) {
+      cli::cli_abort("{.arg perl} not supported when {.arg match} is a stringr pattern")
```
lionel-, line 156:
```suggestion
      cli::cli_abort("{.arg perl} not supported when {.arg match} is a stringr pattern.")
```

R/helpers-pattern.R
```diff
@@ -144,7 +146,23 @@ matches <- function(match,
...
+    }
+    if (length(match) > 1) {
+      cli::cli_abort("stringr patterns must be length 1")
```
lionel-, line 159:
```suggestion
      cli::cli_abort("stringr patterns must be length 1.")
```

lionel-: Could you use `flat_map_int()` as with `grep_vars` here?

lionel- (approved):

R/helpers-pattern.R
hadley: No, because stringr doesn't provide a `[` or `[[` method so all the attributes are lost 😬

lionel-: Ah got it. It would make sense to mention this in a comment.

Not important but we forgot about using the `{.pkg }` style here.

hadley: I'll just commit those changes direct to main.

PR #276, opened 2022-08-10
Match all_of() implementation to any_of()

### LINKED ISSUE #269: Out of context `all_of()` should error
```r
tidyselect::all_of("x")
#> [1] "x"

tidyselect::any_of("x")
#> Error:
#> ! `any_of()` must be used within a *selecting* function.
#> ℹ See <https://tidyselect.r-lib.org/reference/faq-selection-context.html>.
```

hadley:
Fixes #269

### DIFF
--- a/tests/testthat/_snaps/eval-select.md
+++ b/tests/testthat/_snaps/eval-select.md
@@ -4,8 +4,7 @@
       (expect_error(select_loc(mtcars, f())))
     Output
       <error/rlang_error>
-      Error in `select_loc()`:
-      Caused by error in `f()`:
+      Error in `f()`:
       ! foo
 
 # predicate outputs are type-checked
@@ -56,6 +55,7 @@
     Output
       <error/rlang_error>
       Error in `select_loc()`:
+      ! Problem while evaluating `1 + ""`.
       Caused by error in `1 + ""`:
       ! non-numeric argument to binary operator
     Code
--- a/tests/testthat/_snaps/eval-select.md
+++ b/tests/testthat/_snaps/eval-select.md
@@ -4,7 +4,8 @@
       (expect_error(select_loc(mtcars, f())))
     Output
       <error/rlang_error>
-      Error in `f()`:
+      Error in `select_loc()`:
+      Caused by error in `f()`:
       ! foo
 
 # predicate outputs are type-checked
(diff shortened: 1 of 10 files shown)


hadley:
@lionel- any idea why R-devel generates a different parent error?

hadley:
One simple way to fix the issue is to remove the error changing in `eval_context()`. I think that mildly improves the error in the common case of a helper erroring on bad input, at the cost of mildly worsening the error if the user supplies a non-helper that errors.

tests/testthat/test-helpers-vector.R
```diff
@@ -51,10 +38,21 @@ test_that("any_of() is lax", {
 })
 
 test_that("all_of() and any_of() check their inputs", {
-  expect_error(select_loc(letters2, all_of(NA)), "missing")
-  expect_error(select_loc(letters2, any_of(NA)), "missing")
-  expect_error(select_loc(letters2, all_of(na_chr)), "missing")
-  expect_error(select_loc(letters2, any_of(na_chr)), "missing")
-  expect_error(select_loc(letters2, all_of(TRUE)), class = "vctrs_error_subscript_type")
-  expect_error(select_loc(letters2, any_of(TRUE)), class = "vctrs_error_subscript_type")
+  expect_snapshot(error = TRUE, {
```
lionel-, line 41: Could you use `expect_error()` within `expect_snapshot()` to avoid weakening these tests until r-lib/testthat#1667 is solved please?

lionel- (approved):
IIUC you need to install rlang 1.0.4 locally and make tidyselect depend on it so that version is installed on CI. Then the snapshots should be consistent across platforms.

hadley:
Still seeing the failure.

lionel-:
@hadley I'll have to take a deeper look. My local install behaves like r-devel on CI :/

tests/testthat/_snaps/eval-select.md
```diff
@@ -4,8 +4,7 @@
       (expect_error(select_loc(mtcars, f())))
     Output
       <error/rlang_error>
-      Error in `select_loc()`:
-      Caused by error in `f()`:
+      Error in `f()`:
```
lionel-, line 7: Can we keep the chained error here? Chained error support is one the main feature of this release so I'd like to keep them.

PR #277, opened 2022-08-10
Pass error_arg and error_call in vars_pull()

### LINKED ISSUE #258: `vars_pull()` needs arg and call arguments

hadley:
Fixes #258

Not sure that I got `error_arg` right

### DIFF
--- a/R/vars-pull.R
+++ b/R/vars-pull.R
@@ -32,7 +32,7 @@
 #' # You can unquote variables:
 #' var <- 10
 #' vars_pull(letters, !!var)
-vars_pull <- function(vars, var = -1, error_call = caller_env()) {
+vars_pull <- function(vars, var = -1, error_call = caller_env(), error_arg = caller_arg(var)) {
   expr <- enquo(var)
   n <- length(vars)
 
@@ -42,7 +42,7 @@ vars_pull <- function(vars, var = -1, error_call = caller_env()) {
     call = error_call,
     eval_expr = expr
   )
-  loc <- pull_as_location2(loc, n, vars)
+  loc <- pull_as_location2(loc, n, vars, error_arg = error_arg, error_call = error_call)
 
   if (loc < 0L) {
     loc <- n + 1L + loc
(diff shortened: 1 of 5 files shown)


hadley:
Probably should be `error_arg = "var"`? Current code gives:

```R
mtcars %>% pull(all_of(NULL))
#> Error in `pull()`:
#> ! Must extract column with a single valid subscript.
#> ✖ Subscript `!!enquo(var)` has size 0 but must be size 1
```

R/vars-pull.R
```diff
@@ -32,17 +32,21 @@
 #' # You can unquote variables:
 #' var <- 10
 #' vars_pull(letters, !!var)
-vars_pull <- function(vars, var = -1, error_call = caller_env()) {
+vars_pull <- function(vars, var = -1, error_call = caller_env(), error_arg = "var") {
   expr <- enquo(var)
+  if (quo_is_missing(expr)) {
+    expr <- -1
+  }
+
```
lionel-, line 40: Would this give better results?


```r
error_arg <- error_arg %||% as_label(expr)
```

R/vars-pull.R
```diff
@@ -51,23 +55,36 @@ vars_pull <- function(vars, var = -1, error_call = caller_env()) {
...
   with_subscript_errors(type = "pull", {
-    i <- vctrs::vec_as_subscript2(i, arg = "var", logical = "error")
+    i <- vctrs::vec_as_subscript2(i,
+      logical = "error",
+      arg = error_arg,
+      call = error_call
+    )
+
+    if (length(i) != 1) {
+      abort(
```
lionel-, line 67: Needs `cli_abort()` with a `.arg` style class.

R/vars-pull.R
```diff
@@ -51,23 +55,36 @@ vars_pull <- function(vars, var = -1, error_call = caller_env()) {
...
+        glue("`{error_arg}` must select exactly one column"),
```
lionel-, line 68: Missing full stop.

tests/testthat/_snaps/vars-pull.md
```diff
@@ -14,7 +23,7 @@
     Code
       vars_pull(letters, 0)
     Condition
-      Error in `pull_as_location2()`:
+      Error:
```
lionel-, line 26: 👍

R/vars-pull.R
lionel-: If not, `error_arg = ""` is probably better. This should be a value supported by vctrs.

hadley: The first option doesn't look good, and the second option requires more special casing of the manual abort in `pull_as_location2()`. Given that `pull()` is the primary use of this function, and it's argument is called `var`, I think it's a reasonable tradeoff. We can do better in `eval_pull()`.

PR #278, opened 2022-08-10
Improve `where()` errors

### LINKED ISSUE #236: where() with NA values
The `where()` function produces relatively uninformative errors when functions output `NA` values. It would be useful for the user if this was checked.

In this example, the function, `~any(str_detect(.x,","))` returns `TRUE` and `FALSE` values, but also returns `NA` values, and so the below throws an uninformative error:
```
gss_cat %>%
  select(where(~any(str_detect(.x,","))))
Error: `where()` must be used with functions that return `TRUE` or `FALSE`.
```

This can be rectified using `na.rm=T` in the `any()` function, as below:
```
gss_cat %>%
  select(where(~any(str_detect(.x,","),na.rm=T)))
```

I believe that having this throw an error is important, as it draw attention to the `NA` values, but since all the values are `logical`, the error should be more informative.

This is also in contrast to the `select_if()` version of this call:
```
gss_cat %>%
  select_if(~any(str_detect(.,",")))
```

which treats `NA` values the same as `FALSE`

hadley:
Fixes #236

### DIFF
--- a/R/helpers-where.R
+++ b/R/helpers-where.R
@@ -67,7 +67,7 @@ where <- function(fn) {
     out <- predicate(x, ...)
 
     if (!is_bool(out)) {
-      abort("`where()` must be used with functions that return `TRUE` or `FALSE`.")
+      abort(glue("Function in `where()` must return a single `TRUE` or `FALSE`, not {obj_type_friendly(out)}."))
     }
 
     out
--- a/R/helpers-where.R
+++ b/R/helpers-where.R
@@ -67,7 +67,7 @@ where <- function(fn) {
     out <- predicate(x, ...)
 
     if (!is_bool(out)) {
-      abort(glue("Function in `where()` must return a single `TRUE` or `FALSE`, not {obj_type_friendly(out)}."))
+      abort(glue("Predicate function must return a single `TRUE` or `FALSE`, not {obj_type_friendly(out)}."), call = quote(where()))
     }
 
     out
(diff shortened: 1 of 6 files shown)


R/helpers-where.R
```diff
@@ -67,7 +67,11 @@ where <- function(fn) {
     out <- predicate(x, ...)
 
     if (!is_bool(out)) {
-      abort("`where()` must be used with functions that return `TRUE` or `FALSE`.")
+      not <- obj_type_friendly(out)
+      abort(
+        glue("Predicate must return a single `TRUE` or `FALSE`, not {not}."),
+        call = quote(where())
```
hadley, line 73: I think this is correct because `where()` is no longer on the stack, and while this is technically lying about where the error comes from, it's more informative for the user.

lionel-:
Should we make it a chained error?

```
Error in `select_loc()`:"
! Problem while evaluating `where(~ NA)`.
Caused by error:
! Predicate must return `TRUE` or `FALSE` not `NA`.
```

This would have the advantage of showing the exact `where()` expression that is failing, which could be useful when several `where()` expressions are used.

hadley:
Hmmm, I don't think error wrapping helps because `as_indices_sel_impl` doesn't have the quosure. I'll have another look tomorrow.

hadley:
I had a bit of a look and I don't see any obvious way to do better. I don't think it's worth spending too much more time here, but it does now provide the full call to `abort()` so we'll get more context if we improve the heuristics for displaying the call in rlang.

lionel- (approved):

PR #279, opened 2022-08-11
Better error for any_of()

### LINKED ISSUE #241: add example of probable source of error to error message
Just a very minor proposal to add to the error message at https://github.com/r-lib/rlang/blob/1a53bd085c8e0ed41dcc56cce334dde926da7067/R/dots-ellipsis.R#L169

"These dots only exist to allow future extensions and should be empty."
"Did you misspecify an argument?"

It seems a common way to reach this error is forgetting to use c() e.g. I was using any_of("a","b","c") when it should have been any_of(c("a","b","c")) 

Would it be acceptable to add "Did you misspecify an argument? e.g. forget to use c()"?" to this error message? I think it could be helpful for beginners.

hadley:
Fixes #241

### DIFF
--- a/R/helpers-vector.R
+++ b/R/helpers-vector.R
@@ -89,10 +89,10 @@ all_of <- function(x) {
 any_of <- function(x, ..., vars = NULL) {
   vars <- vars %||% peek_vars(fn = "any_of")
   if (!missing(...)) {
-    abort(c(
-      "`...` must be empty.",
-      i = "Did you forget `c()`?",
-      i = 'The expected syntax is `any_of(c("a", "b"))`, not `any_of("a", "b")`'
+    cli::cli_abort(c(
+      "{.arg ...} must be empty.",
+      i = "Did you forget {.code c()}?",
+      i = 'The expected syntax is {.code any_of(c("a", "b"))}, not {.code any_of("a", "b")}'
     ))
   }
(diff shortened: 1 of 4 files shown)


R/helpers-vector.R
```diff
@@ -88,6 +88,12 @@ all_of <- function(x) {
 #' @export
 any_of <- function(x, ..., vars = NULL) {
   vars <- vars %||% peek_vars(fn = "any_of")
-  check_dots_empty()
+  if (!missing(...)) {
+    abort(c(
```
lionel-, line 92: Use `cli_abort()` with `.code` style?

lionel- (approved):

R/helpers-vector.R
hadley: How about I convert them all in a separate PR?

hadley: I see you filed #285 so we're thinking the same way 😄

lionel-: Up to you. Personally for this sort of things I generally convert any code I touch ahead of time, before the bulk changes. Especially for this sort of dull but tricky changes that are so easy to mess up when doing a mass update.

hadley: Yeah, good point.

PR #280, opened 2022-08-11
Always return a named vector from eval_select()

### LINKED ISSUE #220: Return names when renaming is not allowed?
From @DavisVaughan 

```r
tidyselect::eval_select("mpg", mtcars, allow_rename = TRUE)
#> mpg 
#>   1
tidyselect::eval_select("mpg", mtcars, allow_rename = FALSE)
#> [1] 1
```

hadley:
Fixes #220

### DIFF
--- a/R/eval-walk.R
+++ b/R/eval-walk.R
@@ -98,7 +98,7 @@ ensure_named <- function(pos,
     if (is_named(pos)) {
       abort("Can't rename variables in this context.", call = call)
     }
-    return(set_names(pos, NULL))
+    pos
   }
 
   nms <- names(pos) <- names2(pos)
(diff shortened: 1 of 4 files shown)


lionel- (approved):
Hopefully this won't cause revdep issues.

R/eval-walk.R
```diff
@@ -106,7 +106,7 @@ ensure_named <- function(pos,
     if (is_named(pos) && !is_empty(pos)) {
       abort("Can't rename variables in this context.", call = call)
     }
-    return(set_names(pos, NULL))
+    pos
```
lionel-, line 109: oops this might be missing a `return()`.

PR #281, opened 2022-08-11
Improve last_col() input validation

### LINKED ISSUE #211: Conflict between selection of multiple columns in `last_col()` and _R_CHECK_LENGTH_1_LOGIC2_

hadley:
* Check that offset is a scalar
* Use snapshots to test errors
* Add missing periods

Fixes #211

### DIFF
--- a/R/helpers-misc.R
+++ b/R/helpers-misc.R
@@ -62,16 +62,17 @@ everything <- function(vars = NULL) {
 #' @param offset Set it to `n` to select the nth var from the end.
 last_col <- function(offset = 0L, vars = NULL) {
   if (!is_integerish(offset, n = 1)) {
-    abort(glue("`offset` must be a single integer."))
+    not <- obj_type_friendly(offset)
+    cli::cli_abort("{.arg offset} must be a single integer, not {not}.")
   }
 
   vars <- vars %||% peek_vars(fn = "last_col")
   n <- length(vars)
 
   if (offset && n <= offset) {
-    abort(glue("`offset` must be smaller than the number of { plural(vars) }."))
+    cli::cli_abort("{.arg offset} ({offset}) must be smaller than the number of { plural(vars) } ({n}).")
   } else if (n == 0) {
-    abort(glue("Can't select last { singular(vars) } when input is empty."))
+    cli::cli_abort("Can't select last { singular(vars) } when input is empty.")
   } else {
     n - as.integer(offset)
   }
--- a/R/helpers-misc.R
+++ b/R/helpers-misc.R
@@ -61,15 +61,17 @@ everything <- function(vars = NULL) {
 #' @export
 #' @param offset Set it to `n` to select the nth var from the end.
 last_col <- function(offset = 0L, vars = NULL) {
-  stopifnot(is_integerish(offset))
+  if (!is_integerish(offset, n = 1)) {
+    abort(glue("`offset` must be a single integer."))
+  }
 
   vars <- vars %||% peek_vars(fn = "last_col")
   n <- length(vars)
 
   if (offset && n <= offset) {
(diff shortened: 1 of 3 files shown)


R/helpers-misc.R
```diff
@@ -61,15 +61,17 @@ everything <- function(vars = NULL) {
 #' @export
 #' @param offset Set it to `n` to select the nth var from the end.
 last_col <- function(offset = 0L, vars = NULL) {
-  stopifnot(is_integerish(offset))
+  if (!is_integerish(offset, n = 1)) {
+    abort(glue("`offset` must be a single integer."))
```
lionel-, line 65: Here and below: `cli_abort()` with `.arg` style.

lionel-: Can also check for finiteness with `finite = TRUE`. rlang does not export this helper yet but it could be added to `utils.R`:

```r
is_number <- function(x) {
  is_integerish(x, n = 1, finite = TRUE)
}
```

lionel- (approved):

PR #282, opened 2022-08-11
Implement allow_empty

### LINKED ISSUE #252: Option to `eval_select()` to disallow empty selections
i.e.

``` r
library(tidyselect)
library(rlang)

df <- data.frame(x = 1)

# In some cases, I'd like this to be an error
eval_select(expr(starts_with("foo")), df)
#> named integer(0)
```

In tidyr, we have 3 places where we check that the result of `eval_select()` selects "at least 1 column", like
https://github.com/tidyverse/tidyr/pull/1241/files#diff-082bcbb672e069ebb562fb3b877b4d9457f867d89ba382224e1041ce90a89ff4R308

I like that this is already an error by default:

``` r
library(tidyselect)
library(rlang)

df <- data.frame(x = 1)

eval_select(expr(y), df)
#> Error: Can't subset columns that don't exist.
#> x Column `y` doesn't exist.
```

but I think it might make sense to have an ultra strict mode as well that can't ever result in an empty selection

hadley:
Fixes #252

Do you think I need to add an `error_arg` to these errors?

### DIFF
--- a/R/eval-select.R
+++ b/R/eval-select.R
@@ -36,8 +36,8 @@
 #' @param allow_rename If `TRUE` (the default), the renaming syntax
 #'   `c(foo = bar)` is allowed. If `FALSE`, it causes an error. This
 #'   is useful to implement purely selective behaviour.
-#' @param allow_empty If `TRUE` (default), its ok for `expr` to result
-#'   in no variables. If `FALSE`, will error if `expr` yields an empty
+#' @param allow_empty If `TRUE` (the default), it is ok for `expr` to result
+#'   in an empty selection. If `FALSE`, will error if `expr` yields an empty
 #'   selection.
 #' @inheritParams rlang::args_dots_empty
 #'
--- a/R/eval-select.R
+++ b/R/eval-select.R
@@ -36,6 +36,9 @@
 #' @param allow_rename If `TRUE` (the default), the renaming syntax
 #'   `c(foo = bar)` is allowed. If `FALSE`, it causes an error. This
 #'   is useful to implement purely selective behaviour.
+#' @param allow_empty If `TRUE` (default), its ok for `expr` to result
+#'   in no variables. If `FALSE`, will error if `expr` yields an empty
+#'   selection.
 #' @inheritParams rlang::args_dots_empty
 #'
 #' @return A named vector of numeric locations, one for each of the
@@ -122,6 +125,7 @@ eval_select <- function(expr,
                         strict = TRUE,
                         name_spec = NULL,
                         allow_rename = TRUE,
+                        allow_empty = TRUE,
                         error_call = caller_env()) {
   check_dots_empty()
   eval_select_impl(
@@ -133,6 +137,7 @@ eval_select <- function(expr,
     strict = strict,
     name_spec = name_spec,
     allow_rename = allow_rename,
+    allow_empty = allow_empty,
     error_call = error_call,
   )
 }
@@ -146,6 +151,7 @@ eval_select_impl <- function(x,
                              name_spec = NULL,
                              uniquely_named = NULL,
                              allow_rename = TRUE,
+                             allow_empty = TRUE,
                              type = "select",
                              error_call = caller_env()) {
   if (!is_null(x)) {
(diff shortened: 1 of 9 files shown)


R/eval-select.R
```diff
@@ -36,6 +36,9 @@
 #' @param allow_rename If `TRUE` (the default), the renaming syntax
 #'   `c(foo = bar)` is allowed. If `FALSE`, it causes an error. This
 #'   is useful to implement purely selective behaviour.
+#' @param allow_empty If `TRUE` (default), its ok for `expr` to result
+#'   in no variables. If `FALSE`, will error if `expr` yields an empty
+#'   selection.
```
DavisVaughan, line 41:
```suggestion
#' @param allow_empty If `TRUE` (the default), it is ok for `expr` to result
#'   in an empty selection. If `FALSE`, will error if `expr` yields an empty
#'   selection.
```

Just trying to be consistent with `allow_rename`

R/eval-walk.R
```diff
@@ -101,6 +107,8 @@ ensure_named <- function(pos,
     return(set_names(pos, NULL))
   }
 
+  check_empty(pos, allow_empty, call = call)
+
```
DavisVaughan, line 111: Does the check need to be before this early return?

R/eval-walk.R
```diff
@@ -113,6 +121,12 @@ ensure_named <- function(pos,
   pos
 }
 
+check_empty <- function(x, allow_empty = TRUE, call = caller_env()) {
+  if (!allow_empty && length(x) == 0) {
+    abort("Must select at least one item", call = call)
```
DavisVaughan, line 126:
```suggestion
    abort("Must select at least one item.", call = call)
```

Adding the full stop

An alternative error that is similar to the `allow_rename` error is something like:

```
Can't generate an empty selection in this context.
```

DavisVaughan (approved):

R/eval-walk.R
hadley: The seems a bit vague to me — for the tidyr cases, I'd prefer (e.g.) `` `values_from` must select at least one item. ``

DavisVaughan: That seems fine.

So maybe an error arg of some kind would be useful. I think it would probably be useful in a few places though, not just with `allow_empty`, so maybe it is worth doing that in a separate PR

R/eval-walk.R
```diff
@@ -113,6 +121,12 @@ ensure_named <- function(pos,
   pos
 }
 
+check_empty <- function(x, allow_empty = TRUE, call = caller_env()) {
```
lionel-, line 124: Nit: Though it reduces some repetition here, it seems unusual for a `check_` function to take a boolean variable disabling the check. I think the general pattern is more like

```r
if (check) {
  check_foo(x)
}
```

R/eval-walk.R
```diff
@@ -83,16 +87,18 @@ vars_select_eval <- function(vars,
...
 }
 
 ensure_named <- function(pos,
                          vars,
-                         uniquely_named,
-                         allow_rename,
+                         uniquely_named = FALSE,
+                         allow_rename = TRUE,
+                         allow_empty = TRUE,
                          call) {
```
lionel-, line 102: To improve the snapshot:

```suggestion
                         call = caller_env()) {
```

tests/testthat/_snaps/eval-walk.md
```diff
@@ -168,3 +168,11 @@
       Error in `select_loc()`:
       ! The RHS of `.data$rhs` must be a symbol.
 
+# can forbid empty selection
+
+    Code
+      ensure_named(integer(), allow_empty = FALSE)
+    Condition
+      Error in `check_empty()`:
```
lionel-, line 176: I don't understand why it's showing `check_empty()` here.

lionel- (approved):

R/eval-walk.R
hadley: Good catch, thanks.

tests/testthat/_snaps/eval-walk.md
hadley: I think because `call` was `missing` previously

PR #283, opened 2022-08-11
Preserve order of `include`d variables

### LINKED ISSUE #224: `eval_select()` forces variables in `include` to the front
Variables specified with `include` are forced to the front even when they appear later in the selection. This lead to an issue in https://github.com/tidyverse/dbplyr/issues/568

``` r
library(tidyselect)

eval_select(
  rlang::expr(c(a, b)), data.frame(a = 1, b = 1),
  include = "b"
)
#> b a 
#> 2 1
```

<sup>Created on 2020-12-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
Fixes #224


R/eval-select.R
```diff
@@ -162,7 +162,7 @@ eval_select_impl <- function(x,
   local_data(x)
 
   if (length(include)) {
-    expr <- quo(all_of(include) | !!expr)
+    expr <- quo((all_of(include) & !(!!expr)) | !!expr)
```
lionel-, line 165: I think this introduces double evaluation? Does the following work?

```suggestion
    expr <- quo({
      `_tmp` <- !!expr
      (all_of(include) & !all_of(`_tmp`)) | `_tmp`
    })
```

I would not be surprised if it doesn't work. Perhaps we need recursive calls to `eval_select()` here to solve this?

lionel-: Will need a test for double evaluation.

hadley: Do you think double evaluation is troublesome? I'd expect that most selections would be fast and stateless.

lionel-: I'd expect the same, but double evaluation is still a bug I think.

lionel-: The main way in which trouble could manifest is through predicate functions (side effects like warnings, non trivial times with wide and long data frames).

I can also take a look if you'd like. Worst case we'll just leave a comment about the issue.

hadley: Yeah, why don't you finish this one off.

R/eval-select.R
```diff
@@ -162,7 +162,21 @@ eval_select_impl <- function(x,
   local_data(x)
 
   if (length(include)) {
-    expr <- quo((all_of(include) & !(!!expr)) | !!expr)
+    # Recurse to transform `expr` into a vector of locations. This
```
lionel-, line 171: This sort of recursion seems risky because, if new arguments are added, it's easy to forget to add them here.

This encourages me to try and find a solution for https://github.com/r-lib/rlang/issues/1363.

lionel-:
@hadley Ready for review.

R/eval-select.R
hadley: Before we go even further down this route, we should consider if it would be easier if `eval_select_impl` handed `include` and `exclude`.

tests/testthat/test-eval-select.R
```diff
@@ -81,6 +112,8 @@ test_that("predicate outputs are type-checked", {
 })
 
 test_that("eval_select() produces correct backtraces", {
+  skip_if(interactive()) # backtraces look different
```
lionel-, line 115: This is going to make it hard to regenerate / monitor backtraces. What exactly was different?

lionel-: In ESS `devtools::test()` runs in the same interactive session. I'll take a look to make the test stable.

hadley: If I remove from this PR, is it good to merge?

lionel-: yes thanks!

PR #288, opened 2022-08-12
Focus on Boolean algebra

### LINKED ISSUE #203: Technical description should have more clear example about the difference between ! and -
As [the doc](https://tidyselect.r-lib.org/articles/syntax.html#dots-c-and-unary-) describes

> In this case, unary `-` is syntax for set complement. Unary `-` and `!` are equivalent:

we might think they are completely the same. But, they are NOT equivalent when there's some preceding selection. I think the section has some example to highlight the difference like this (we might want to replace the use of `iris` before doing this, though):

``` r
library(tidyselect)
library(magrittr)

select_loc <- function(data, ...) {
  eval_select(rlang::expr(c(...)), data)
}

iris %>% select_loc(c(starts_with("Sepal"), !Sepal.Length))
#> Sepal.Length  Sepal.Width Petal.Length  Petal.Width      Species 
#>            1            2            3            4            5

iris %>% select_loc(c(starts_with("Sepal"), -Sepal.Length))
#> Sepal.Width 
#>           2
```

<sup>Created on 2020-07-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
Fixes #203

I removed most of the discussion about `-`. I don't think it's worth precisely defining the semantics of a syntax that we no longer recommend.

### DIFF
--- a/vignettes/syntax.Rmd
+++ b/vignettes/syntax.Rmd
@@ -61,10 +61,13 @@ match, they are treated as different elements in order to allow
 renaming a variable to multiple names (see section on Renaming
 variables).
 
-The syntax of tidyselect is generally designed for set combination.
-For instance, `c(foo(), bar())` represents the union of the variables
-in `foo()` and those in `bar()`.
-
+Today, the syntax of tidyselect is generally designed around Boolean algebra,
+i.e. we recommend writing `starts_with("a") & !ends_with("z")`. Earlier
+versions of tidyselect had more of a flavour of set operations, so that 
+you'd write `starts_with("a") - ends_with("b")`. While the set operations are
+still supported, and how tidyselect represents variables internally, we no 
+longer recommend them, because the Boolean algebra is easy for people to
+understand.
 
 ### Bare names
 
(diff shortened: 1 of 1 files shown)


vignettes/syntax.Rmd
```diff
@@ -61,10 +61,13 @@ match, they are treated as different elements in order to allow
...
 
-The syntax of tidyselect is generally designed for set combination.
-For instance, `c(foo(), bar())` represents the union of the variables
-in `foo()` and those in `bar()`.
-
+Today, the syntax of tidyselect is generally designed around Boolean algebra,
+i.e. we recommend writing `starts_with("a") & !ends_with("z")`. Earlier
+versions of tidyselect had more of a flavour of set operations, so that 
+you'd write `starts_with("a") - ends_with("b")`. While the set operations are
+still supported, and how tidyselect represents variables internally, we no
```
lionel-, line 68:
```suggestion
still supported, and is how tidyselect represents variables internally, we no 
```

vignettes/syntax.Rmd
```diff
@@ -61,10 +61,13 @@ match, they are treated as different elements in order to allow
...
-The syntax of tidyselect is generally designed for set combination.
-For instance, `c(foo(), bar())` represents the union of the variables
-in `foo()` and those in `bar()`.
-
+Today, the syntax of tidyselect is generally designed around Boolean algebra,
+i.e. we recommend writing `starts_with("a") & !ends_with("z")`. Earlier
+versions of tidyselect had more of a flavour of set operations, so that 
+you'd write `starts_with("a") - ends_with("b")`. While the set operations are
+still supported, and how tidyselect represents variables internally, we no 
+longer recommend them, because the Boolean algebra is easy for people to
```
lionel-, line 69:
```suggestion
longer recommend them because Boolean algebra is easy for people to
```

lionel- (approved):

PR #290, opened 2022-08-12
Give advice on reusable selection helpers

### LINKED ISSUE #270: read_csv(col_select = dplyr::select() variable) fails at variable declaration time
What I expected to work:
```R
columnsOfInterest = c("Year", "ID1", starts_with("Tmin"))
bind_rows(read_csv("file1.csv", col_select = columnsOfInterest),
          read_csv("file2.csv", col_select = columnsOfInterest))
```

What actually happened:
```R
Error:
! `starts_with()` must be used within a *selecting* function.
ℹ See <https://tidyselect.r-lib.org/reference/faq-selection-context.html>.
Run `rlang::last_error()` to see where the error occurred.
> rlang::last_error()
<error/rlang_error>
Error:
! `starts_with()` must be used within a *selecting* function.
ℹ See <https://tidyselect.r-lib.org/reference/faq-selection-context.html>.
---
Backtrace:
 1. tidyselect::starts_with("Tmin")
 3. tidyselect::peek_vars(fn = "starts_with")
Run `rlang::last_trace()` to see the full context.
> rlang::last_trace()
<error/rlang_error>
Error:
! `starts_with()` must be used within a *selecting* function.
ℹ See <https://tidyselect.r-lib.org/reference/faq-selection-context.html>.
---
Backtrace:
    ▆
 1. └─tidyselect::starts_with("Tmin")
 2.   ├─vars %||% peek_vars(fn = "starts_with")
 3.   └─tidyselect::peek_vars(fn = "starts_with")
 4.     └─rlang::abort(msg, call = NULL)
```

What the best workaround appears to be:
```R
bind_rows(read_csv("file1.csv", col_select = c("Year", "ID1", starts_with("Tmin")),
          read_csv("file2.csv", col_select = c("Year", "ID1", starts_with("Tmin")))
```

It seems strange to force users loading a bunch of similar .csv files to repeatedly indicate `col_select` rather than declaring it once. As with many such things it's not a big deal once you figure out what to do, though I did find the linked FAQ and other readr and dplyr documentation to have no relevant content.

It'd be nice to be able to write more proper code instead of relying on find/replace to maintain many hardcoded instances of the same thing. This is, as usual, a trivialized repex—the actual `c()` statement is considerably more complex—and I certainly wasn't expecting which line of code a `c()` was declared in would influence whether the declaration was valid.

I'm putting this issue in tidyselect since it seems like this might be worth a look from the context of dplyr::select() mini-language design it appears this is the repo which covers that. From the backtrace it seems `peek_vars()` might be prematurely positioned.

hadley:
Fixes #270

After writing this, I'm not sure this is quite the right place, but we definitely need this advice somewhere.

### DIFF
--- a/man/faq/selection-context.Rmd
+++ b/man/faq/selection-context.Rmd
@@ -3,12 +3,8 @@
 ```
 
 Functions like `starts_with()`, `contains()` or `matches()` are
-__selection helpers__ that only work in a selection context.
-
-Examples of valid selection contexts are:
-
-- Inside `dplyr::select()`.
-- The `cols` argument of `tidyr::pivot_longer()`.
+__selection helpers__ that only work in a selection context, e.g.
+`dplyr::select()` or the `cols` argument of `tidyr::pivot_longer()`.
 
 Using a selection helper anywhere else results in an error:
 
@@ -20,6 +16,24 @@ mtcars[contains("foo")]
 subset(mtcars, select = matches("foo"))
 ```
 
-If you see this error, you've probably used a selection helper in the
+If you see this error, you may have used a selection helper in the
 wrong place, possibly as the result of a typo (e.g. misplaced comma or
-wrong argument name).
+wrong argument name). Alternatively, you may be deliberately trying
+to reduce duplication in your code by extracting out a selection into
+a variable:
+
+```{r, error = TRUE}
+my_vars <- c(name, species, ends_with("color"))
+```
+
+To make this work you'll need to do two things:
+
+* Wrap the whole thing in a function
+* Use `any_of()` or `all_of()` instead of bare variable names
+
+```{r}
+my_vars <- function() {
+  c(any_of(c("name", "species")), ends_with("color"))
+}
+dplyr::select(starwars, my_vars())
+```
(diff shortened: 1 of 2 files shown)


man/faq/selection-context.Rmd
```diff
@@ -20,6 +16,24 @@ mtcars[contains("foo")]
...
+a variable:
+
+```{r, error = TRUE}
+my_vars <- c(name, species, ends_with("color"))
+```
+
+To make this work you'll need to do two things:
+
+* Wrap the whole thing in a function
+* Use `any_of()` or `all_of()` instead of bare variable names
```
lionel-, line 32: Does `all_of()` now return a vector of positions after the recent changes? If yes, do we have a unit test for the combination of `all_of()` and helpers like `starts_with()`? If not, we should keep #270 around, or open a new issue.

lionel-: Should we add an info bullet pointing to this page in the out-of-context error?

hadley: I think nothing has changed with `all_of()`, except that (e.g) `all_of(1)` is now coerced to an integer vector.

And this is the Rd topic that `peek_vars()` points to 😄

PR #291, opened 2022-08-12
Use help hyperlink in peek_vars()

### LINKED ISSUE #289: peek_vars() error should generate link to docs
i.e. if RStudio hyperlinks are available, it should link to `?tidyselect::'faq-selection-context'`.

hadley:
Fixes #289

### DIFF
--- a/R/vars.R
+++ b/R/vars.R
@@ -14,10 +14,17 @@ peeker <- function(what) {
         fn <- glue::glue("`{fn}()`")
       }
 
+      if (cli::ansi_has_hyperlink_support()) {
+        topic <- "tidyselect::faq-selection-context"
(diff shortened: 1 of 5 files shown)


R/vars.R
```diff
@@ -14,10 +14,17 @@ peeker <- function(what) {
         fn <- glue::glue("`{fn}()`")
       }
 
+      if (is_interactive() && cli::ansi_has_hyperlink_support()) {
```
lionel-, line 17: Should you check for `cli::num_ansi_colors() != 1` as well to fix the snapshots?

Technically we should also use the feature guard but this will eventually be handled by cli.

lionel- (approved):

R/vars.R
hadley: Maybe to get the tests passing it's better to set the other env var too

PR #299, opened 2022-08-17
Add a class to the rename error and rethrow it in `eval_relocate()`

### LINKED ISSUE #298: Throw a classed error when the `allow_rename = FALSE` error is thrown
https://github.com/r-lib/tidyselect/pull/297#discussion_r947906210

And rethrow it from `eval_relocate()` when calling `eval_select()` with `before` and `after`. Improve on the error message by using `before_arg` and `after_arg`

DavisVaughan:
Closes #298 

I didn't add a class to the rethrown error. It seems like the class is just meant for internal use, and I don't see us needing to ever rethrow the already rethrown error, so I didn't think it needed a class.

### DIFF
--- a/R/eval-relocate.R
+++ b/R/eval-relocate.R
@@ -107,12 +107,16 @@ eval_relocate <- function(expr,
   }
 
   if (has_before) {
-    where <- eval_select(
-      expr = before,
-      data = data,
-      env = env,
-      error_call = error_call,
-      allow_rename = FALSE
+    where <- with_rename_errors(
+      eval_select(
+        expr = before,
+        data = data,
+        env = env,
+        error_call = error_call,
+        allow_rename = FALSE
+      ),
+      arg = before_arg,
+      error_call = error_call
     )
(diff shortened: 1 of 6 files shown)


R/eval-relocate.R
```diff
@@ -161,3 +169,15 @@ eval_relocate <- function(expr,
 
   sel
 }
+
+with_rename_errors <- function(expr, arg, error_call) {
+  tryCatch(
```
lionel-, line 174:
```suggestion
  try_fetch(
```

To get a more complete full backtrace and a more useful `recover()`.

lionel- (approved):

R/eval-relocate.R
DavisVaughan: Do you want me to also change to `try_fetch()` in `with_subscript_errors()` then? I'll do it in another PR

lionel-: Sounds good, thanks

PR #304, opened 2022-08-24
Simplify `starts_with()` and `ends_with()`

cphaarmeyer:


hadley:
Unfortunately we can't use these functions because we also support older versions of R.

PR #307, opened 2022-08-30
Clarify selection helpers & regexes

### LINKED ISSUE #306: Documentation: clarity around regular expressions in ?starts_with
I went to https://tidyselect.r-lib.org/reference/starts_with.html to try and understand if `starts_with()` accepts regular expressions, and came away unclear.

The `match` parameter is shared across all the functions and described as:

> A character vector.

There is a `perl` argument on the page that's only accepted by by `matches`, so I can intuit that `matches` is the only regex-friendly member of the page, but this relies a bit on my knowledge of arcana.

There's also this:

https://github.com/r-lib/tidyselect/blob/0bc35e7fb75d79e668e6674835caa5f2692be60f/R/helpers-pattern.R#L63-L64

But it's not clear what "these" refers to here. Maybe "The above" would be better, but best would be to name all the non-regex functions explicitly.

MichaelChirico:
Closes #306.

### DIFF
--- a/R/helpers-pattern.R
+++ b/R/helpers-pattern.R
@@ -5,8 +5,8 @@
 #' These [selection helpers][language] match variables according
 #' to a given pattern.
 #'
-#' * [starts_with()]: Starts with a prefix.
-#' * [ends_with()]: Ends with a suffix.
+#' * [starts_with()]: Starts with an exact prefix.
+#' * [ends_with()]: Ends with an exact suffix.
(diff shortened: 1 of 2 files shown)


MichaelChirico:
Someone please help to document, I am unable to install all the required dev versions locally 😬

hadley:
Sorry, dev versions are in high flux at the moment!

PR #309, opened 2022-09-13
Implement generic interface

### LINKED ISSUE #242: Support for lazy tibbles
Lazy tibbles currently have two issues with `eval_select()`:

1. They are not a vector so one has to simulate a "dummy" tibble with the appropriate column names, e.g with `dtplyr:::simulate_vars()`.
2. The column types are not generally known which is why `allow_predicates` would make sense: #226

The second point is still a bit annoying when trying to do some general work in a generic, e.g. in `tidyr::nest()` unnamed dots are tried to be named

```r
tidyr::nest <- function (.data, ..., .names_sep = NULL, .key = deprecated()) {
  cols <- enquos(...)
  if (any(names2(cols) == "")) {
    col_names <- names(tidyselect::eval_select(expr(c(...)), .data))
    ...
  }
  UseMethod("nest")
}
```

So, it would be necessary to check the type to figure out whether `allow_predicate` should be `TRUE` or `FALSE`. It might make sense if `tidyselect()` could handle this.

Also see https://github.com/tidyverse/tidyr/issues/1134

hadley:
Fixes #242


R/eval-relocate.R
```diff
@@ -83,6 +89,7 @@ eval_relocate <- function(expr,
     name_spec = name_spec,
     allow_rename = allow_rename,
     allow_empty = allow_empty,
+    allow_predicates = allow_predicates,
     type = "relocate",
     error_call = error_call
   )
```
DavisVaughan, line 95: I think the `eval_select()` calls for `.before` and `.after` (further below) need to know about this `allow_predicates` argument somehow.

Assuming `data <- tidyselect_data_proxy(data)` returns an R data frame if it was original a lazy df, then the `eval_select()` call for `before` and `after` would look like they _do_ support predicates, but they really shouldn't.

R/proxy.R
```diff
@@ -0,0 +1,25 @@
...
+#'    `tidyselect_data_proxy()` doesn't return actual types.
+#'
+#' @param x A data-frame like object passed to `[eval_select()]`,
+#'   `eval_rename()`, and friends.
+#' @export
+tidyselect_data_proxy <- function(x) {
+  UseMethod("tidyselect_data_proxy")
+}
+#' @export
+tidyselect_data_proxy.default <- function(x) x
```
DavisVaughan, line 14:
```suggestion
tidyselect_data_proxy.default <- function(x) {
  x
}
```

R/proxy.R
```diff
@@ -0,0 +1,25 @@
+#' tidyselect methods for custom types
+#'
+#' * `tidyselect_data_proxy()` should return a 0-row data frame.
+#' * `tidyselect_data_supports_predicates()` should return `FALSE` if
+#'    `tidyselect_data_proxy()` doesn't return actual types.
+#'
+#' @param x A data-frame like object passed to `[eval_select()]`,
+#'   `eval_rename()`, and friends.
```
DavisVaughan, line 8:
```suggestion
#' @param x A data-frame like object passed to [eval_select()],
#'   [eval_rename()], and friends.
```

R/proxy.R
```diff
@@ -0,0 +1,25 @@
+#' tidyselect methods for custom types
+#'
+#' * `tidyselect_data_proxy()` should return a 0-row data frame.
+#' * `tidyselect_data_supports_predicates()` should return `FALSE` if
+#'    `tidyselect_data_proxy()` doesn't return actual types.
```
DavisVaughan, line 5:
```suggestion
#'   `tidyselect_data_proxy()` doesn't return actual types.
```

tests/testthat/test-proxy.R
```diff
@@ -0,0 +1,22 @@
+test_that("eval_*() respects proxy settings", {
+  foo <- structure(list(), class = "foo")
+  local_bindings(
+    tidyselect_data_proxy.foo = function(x) {
+      data.frame(x = 1, y = 2)
+    },
+    tidyselect_data_supports_predicates.foo = function(x) {
+      FALSE
+    },
+    .env = globalenv()
```
DavisVaughan, line 10:
```suggestion
    }
```

R/eval-relocate.R
```diff
@@ -67,12 +67,18 @@ eval_relocate <- function(expr,
                           name_spec = NULL,
                           allow_rename = TRUE,
                           allow_empty = TRUE,
+                          allow_predicates = NA,
```
DavisVaughan, line 70: I was actually not under the impression that this would be a user facing argument. I thought the generics would be in full control of whether or not predicates are allowed.

DavisVaughan (approved):
Approved if you are certain that `allow_predicates` should be a user facing argument

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # tidyselect (development version)
 
+* New `tidyselect_proxy_data()` and `tidyselect_proxy_supports_predicates()`
```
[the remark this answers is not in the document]
hadley: Any ideas for better names? I have now mistyped them 3+ times.

R/eval-relocate.R
hadley: Yeah, that was the original goal, but then it seemed inconsistent to not also have it at the top level.

NEWS.md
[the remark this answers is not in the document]
DavisVaughan: I'd be fine with dropping `data`

```r
tidyselect_proxy(data)
tidyselect_supports_predicates(data)
```

It feels slightly redundant since `data` is the argument name too.

And I'm not too worried about it ever being ambiguous with anything else in tidyselect

tests/testthat/test-proxy.R
hadley: This is necessary, otherwise the S3 registration stuff doesn't work right.

tests/testthat/test-proxy.R
```diff
@@ -6,7 +6,8 @@ test_that("eval_*() respects proxy settings", {
     },
     tidyselect_data_supports_predicates.foo = function(x) {
       FALSE
-    }
+    },
+    .env = globalenv()
```
DavisVaughan, line 10: Oh sorry, I guess I thought this was `local_methods()`

https://github.com/r-lib/vctrs/blob/e86bb9a4b57459eb56099c0b52af37877a9bf515/tests/testthat/helper-s3.R#L34-L36

R/eval-relocate.R
```diff
@@ -67,12 +67,18 @@ eval_relocate <- function(expr,
...
                           before_arg = "before",
                           after_arg = "after",
                           env = caller_env(),
                           error_call = caller_env()) {
   check_dots_empty()
 
+  if (is.na(allow_predicates)) {
+    allow_predicates <- tidyselect_data_supports_predicates(data)
```
lionel-, line 78: If the frontend doesn't support predicates and `eval_select(allow_predicates = TRUE)` has been invoked, should it be:

- An error?
- The frontend has the last word?

I think it would make sense for the caller of `eval_select()` to turn off predicate support, but not enable it if that isn't possible.

R/proxy.R
```diff
@@ -0,0 +1,27 @@
...
+
+
+#' @rdname tidyselect_data_proxy
+#' @export
+tidyselect_data_supports_predicates <- function(x) {
+  UseMethod("tidyselect_data_supports_predicates")
+}
+#' @export
+tidyselect_data_supports_predicates.default <- function(x) {
+  TRUE
```
lionel-, line 26: Should this be an enum `c("full", "ptype", "none")`?

R/eval-relocate.R
hadley: Oh yeah, it probably should be `allow_predicates <- allow_predicates && tidyselect_data_supports_predicates(data)`

R/eval-relocate.R
```diff
@@ -67,12 +67,18 @@ eval_relocate <- function(expr,
...
                           allow_rename = TRUE,
                           allow_empty = TRUE,
+                          allow_predicates = NA,
                           before_arg = "before",
                           after_arg = "after",
                           env = caller_env(),
                           error_call = caller_env()) {
   check_dots_empty()
 
+  if (is.na(allow_predicates)) {
```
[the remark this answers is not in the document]
DavisVaughan: I also like `NULL` as a default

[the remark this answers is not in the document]
hadley: See below — it should probably be `TRUE` and then we do `allow_predicates <- allow_predicates && tidyselect_data_supports_predicates(data)`

R/eval-walk.R
```diff
@@ -172,18 +175,26 @@ walk_data_tree <- function(expr, data_mask, context_mask, colon = FALSE) {
...
+    allow_predicates = allow_predicates,
     call = error_call
   )
 }
 
-as_indices_sel_impl <- function(x, vars, strict, data = NULL, call) {
+as_indices_sel_impl <- function(x, vars, strict, data = NULL, allow_predicates = TRUE, call) {
   if (is.function(x)) {
+    if (!allow_predicates) {
+      cli::cli_abort(
```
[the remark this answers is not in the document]
hadley: The class idea is good, but feeding through a type summary of the original object is going to be hard because we'd need to capture it before proxying it and pass it down through many levels.

R/proxy.R
```diff
@@ -0,0 +1,27 @@
+#' tidyselect methods for custom types
+#'
+#' * `tidyselect_data_proxy()` should return a 0-row data frame.
```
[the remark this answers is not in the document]
hadley: Ooops, you're right.

We could make this a hierarchy, but we'd need to use some heuristic to figure out if a predicate was a type function or something else. So while it would be a nice improvement to the interface, I don't think it's worth the effort.

R/eval-select.R
```diff
@@ -39,6 +39,10 @@
 #' @param allow_empty If `TRUE` (the default), it is ok for `expr` to result
 #'   in an empty selection. If `FALSE`, will error if `expr` yields an empty
 #'   selection.
+#' @param allow_predicates If `TRUE` (the default), it is ok for `expr` to
+#'   use predicate (i.e. in `where()`). If `FALSE`, will error if `expr` uses a
+#'   predicate. Will automatically be set to `FALSE` if `data` does not
+#'   support predicates (as determined by [tidyselect_data_has_predicates()].
```
DavisVaughan, line 45:
```suggestion
#'   support predicates (as determined by [tidyselect_data_has_predicates()]).
```

R/eval-select.R
```diff
@@ -39,6 +39,10 @@
 #' @param allow_empty If `TRUE` (the default), it is ok for `expr` to result
 #'   in an empty selection. If `FALSE`, will error if `expr` yields an empty
 #'   selection.
+#' @param allow_predicates If `TRUE` (the default), it is ok for `expr` to
+#'   use predicate (i.e. in `where()`). If `FALSE`, will error if `expr` uses a
```
DavisVaughan, line 43:
```suggestion
#'   use predicates (i.e. in `where()`). If `FALSE`, will error if `expr` uses a
```

DavisVaughan (approved):

R/eval-walk.R
```diff
@@ -172,18 +175,27 @@ walk_data_tree <- function(expr, data_mask, context_mask, colon = FALSE) {
...
+        "This tidyselect interface doesn't support predicates.",
+        call = call,
+        class = "tidyselect_predicates_unsupported"
```
lionel-, line 196: We normally use `mypkg_error_` for errors:

```suggestion
        class = "tidyselect_error_predicates_unsupported"
```

R/proxy.R
```diff
@@ -0,0 +1,31 @@
+#' tidyselect methods for custom types
+#'
+#' @description
+#' * `tidyselect_data_proxy()` returns a data frame.
+#' * `tidyselect_data_has_predicates()` returns `TRUE` or `FALSE`
+#'
+#' If your doesn't support predicate functions, return a 0-row data frame
+#' from `tidyselect_data_proxy()` and `FALSE` from
+#' `tidyselect_data_has_predicates()`.
```
lionel-, line 9: 👍

lionel- (approved):

PR #313, opened 2022-09-16
Deprecate `all_of()` instead of erroring

hadley:
This resolved 30 revdep failures.


lionel- (approved):

hadley:
@lionel- can you take a look at my second commit? It turns out this only affects broom.helpers, so alternatively I could fix there.

hadley:
Actually, I think we have to support `all_of()` in `pull()`? Otherwise how do you get `vars <- "x"` to work?

hadley:
And should probably make this a soft deprecation, since it affects a decent number of packages.

lionel- (approved):

PR #314, opened 2022-09-16
Make vars_pull() error if embraced arg is missing

hadley:


R/vars-pull.R
```diff
@@ -35,7 +35,9 @@
 vars_pull <- function(vars, var = -1, error_call = caller_env(), error_arg = "var") {
   expr <- enquo(var)
   if (quo_is_missing(expr)) {
-    expr <- -1
+    # No easy way to determine what var is in parent because it's likely
+    # to be embraced; so don't try and use error_arg + error_call here
+    cli::cli_abort("{.arg var} is absent but must be supplied.")
```
lionel-, line 40: I think this is why we take `error_arg`? What are the failing cases when you supply `var = error_arg, call = error_call`?

hadley: I think this is probably some rlang bug or limitation:

``` r
library(rlang)

vars_pull <- function(vars, var = -1, error_call = caller_env(), error_arg = caller_arg(var)) {
  
  expr <- enquo(var)
  if (quo_is_missing(expr)) {
    cli::cli_abort("{.arg {error_arg}} is missing")
  }
  
}

vars_pull(mtcars)

f <- function(df, var) {
  vars_pull(df, {{ var }})
}
f(mtcars)
#> Error in `vars_pull()`:
#> ! `{ ... }` is missing
```

<sup>Created on 2022-09-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

lionel-: hmm yes. I think we can still pass `error_call`, and maybe we should use a better default for `error_arg`. Either `"var"` or make it optional (which is annoying because we need two variants of error messages).

PR #349, opened 2024-04-01
Fix grammar in warning about bare predicates

matthewjnield:
When a predicate function is used in `select()` to select columns for which a function returns `TRUE`, a warning appears, saying: "Please use wrap predicates in `where()` instead.", which is grammatically incorrect.

To correct the warning's grammar, this commit changes the warning to say: "Please use `where()` to wrap predicate functions instead."


lionel-:
Thanks!

PR #352, opened 2024-07-11
Improve `eval_select()` error chaining

olivroy:
I noticed that there was an issue with backtraces. Turns out a simple forgotten error call caused it.

``` r
# before
library(tidyselect)
f <- function(expr, call = rlang::caller_env()) {
  eval_select(
    expr = {{ expr }},
    data = mtcars
  )
}
f(16)
#> Error in `vars_select_eval()`:
#> ! Can't select columns past the end.
#> ℹ Location 16 doesn't exist.
#> ℹ There are only 11 columns.
```

<sup>Created on 2024-07-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

``` r
# this PR
devtools::load_all(".")
#> ℹ Loading tidyselect
f <- function(expr, call = rlang::caller_env()) {
  eval_select(
    expr = {{ expr }},
    data = mtcars
  )
}
f(16)
#> Error in `f()`:
#> ! Can't select columns past the end.
#> ℹ Location 16 doesn't exist.
#> ℹ There are only 11 columns.
```

<sup>Created on 2024-07-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I don't mind adding a test if you think it is worthwhile.

---

Upstream, I am working on gt.

Here is the nice effect this PR will have on gt when the next tidyselect version is released. (above is the new message, below is with CRAN tidyselect)
image


lionel-:
Thanks!

Issue #333, opened 2023-01-19
Include examples of named `any_of()`/`all_of()`


hadley:
https://github.com/tidyverse/dplyr/issues/6644

mike-duran-mitchell:
First of all, I absolutely appreciate that these examples were added because debugging a large script with the `.groups` message is pure hell without it explicitly showing where it occurred. More examples for deprecations or warnings like this, please!!! Additionally, it'd be nice to make this example a little more clear. For example, my code is:

```
cols_to_select <- c('col1', 'col2', col3')
subset_df <- df %>% select(-cols_to_select)
```

But the warning message doesn't indicate where the `-` sign goes. My suggestion would be just to add it in below so that it's in front of `all_of()` in the below example when the deprecation warning is thrown. Not really the hardest curveball for me to knock out of the park, but I can imagine newer users would feel a little confused about whether to stick the `-` inside or outside of `all_of()`.

```
Warning message:
Using an external vector in selections was deprecated in tidyselect 1.1.0.
ℹ Please use `all_of()` or `any_of()` instead.
  # Was:
  data %>% select(cols_to_select)

  # Now:
  data %>% select(all_of(cols_to_select))

See <https://tidyselect.r-lib.org/reference/faq-external-vector.html>.
This warning is displayed once every 8 hours.
Call `lifecycle::last_lifecycle_warnings()` to see where this warning was generated. 
```

I got it right the first time using `select(-all_of(cols_to_select))` but I also wanted to confirm the function is not agnostic regarding the placement of `-` and it does throw an error:

```
Error in `dplyr::select()`:
! Problem while evaluating `all_of(-cols_to_select)`.
Caused by error in `-cols_to_select`:
! invalid argument to unary operator
Run `rlang::last_trace()` to see where the error occurred.
```

lionel-:
Closed by #356.

PR #356, opened 2024-08-30
Add example of named vector x for all_of() and any_of()

laplasi:
`all_of()` and `any_of()` have special behaviour when `x` is a named vector. See #333. 

(First commit to a project)


lionel-:
Thanks!

Issue #336, opened 2023-02-22
Can we add more context for `Can't rename in this context` errors?


DavisVaughan:
This came up in https://github.com/tidyverse/dplyr/issues/6745

We've started using `allow_rename = FALSE` more often when our interfaces use tidy selection _purely_ for selection (i.e. when we don't allow named inputs for possible renaming), so this error has popped up more often.

I think it would be useful if we could mention the expression causing the issue, like how we do here:

```r
library(tidyselect)
library(rlang)

eval_select(expr(all_of(x)), mtcars)
#> Error:
#> ! Problem while evaluating `all_of(x)`.
#> Caused by error in `as_indices_impl()`:
#> ! object 'x' not found
```

I'm imagining something like this for https://github.com/tidyverse/dplyr/issues/6745:

```r
#> Error in `rename_with()`:
#> ! Problem while evaluating `all_of(x)`.
#> ! Can't rename variables in this context.
```

DavisVaughan:
Another one https://github.com/tidyverse/tidyr/issues/1481

olivroy:
@DavisVaughan I liked your suggestion in https://github.com/r-lib/tidyselect/pull/358#discussion_r1813411720, but I didn't feel confident enough to implement it. If you have time, feel free to revisit!

Issue #327, opened 2022-10-18
More context for errors


hadley:
e.g.

```R
pivot_longer(df, c(z = y))
#> Error in `build_longer_spec()`:
#> ! Can't rename variables in this context.
```

Maybe just needs an `error_arg`?

lionel-:
hmm yes, and also pass the error call?

hadley:
Error call is already there (`build_longer_spec()` is an exported tidyr function so I didn't override the call).

lionel-:
Ah in this case I think the error would be more informative for end users if you passed the call to `build_longer_spec()`. I think we can conceive of it as a backend function that takes `error_call = current_env()` (instead of `caller_env()`) so that callers may borrow its errors but it's still producing the correct call when called directly.

hadley:
I think that's probably a good idea. We should probably talk through (and write up) exactly when it's appropriate for a function to take an `error_call` argument.

lionel-:
@olivroy contributed support for this in https://github.com/r-lib/tidyselect/pull/350 and https://github.com/r-lib/tidyselect/pull/359

PR #358, opened 2024-10-23
Improve error message + tweak function signature respect `error_arg` for `allow_rename = FALSE`

olivroy:
I split the PR up

* Use recent standalone files #360 
* Tweak function signature (from #350 )  and support `error_arg` for `allow_rename = FALSE` #359 
* Use standard snapshot tests with `cnd_class = TRUE` when appropriate. #361 

No rush, but I wondered if you had a date in mind for a cran release?


olivroy:
With 9938eb5, I may have a solution for #327, #336. 

I will test it with dplyr / tidyr to see if it works well.

DavisVaughan:
Would it be possible to make this into multiple much smaller PRs that do exactly 1 thing?
https://code-review.tidyverse.org/author/focused.html

It is much easier for us to confidently review a PR when it does 1 thing. In particular this is 3-4 PRs:
- The `i = "{.arg {error_arg}} can't be renamed."` error bullet
- Updating to `use_standalone()` helpers, and updating type checker usage
- Modernizing snapshot tests
- Changing a function signature

These each deserve their own PR where we can have detailed but focused conversation about each idea!

R/eval-walk.R
```diff
@@ -110,11 +110,22 @@ ensure_named <- function(pos,
...
+      )
+    } else {
+      cli::cli_abort(c(
+        "Can't rename variables in this context.",
+        i = "{.arg {error_arg}} can't be renamed."
+        ),
+        class = "tidyselect:::error_disallowed_rename",
+        call = call
+      )
+    }
```
DavisVaughan, line 127: In https://github.com/r-lib/tidyselect/issues/336 I think I was hoping that we could somehow mention the expression where the rename happens in the error message.

`ensure_named()` is only called by `vars_select_eval()` and that does have the `expr`. So you could pass `expr` through to `ensure_named()` and use `as_label(expr)` on it to get something nice to include in the error message. Like

```r
      expr <- as_label(expr)
      cli::cli_abort(c(
        "Can't rename variables during selection.",
        i = "Rename detected in {.arg {error_arg}}, with expression {.code {expr}}."
        ),
        class = "tidyselect:::error_disallowed_rename",
        call = call
      )
```

For something like `c(all_of(x), y)` if the rename happens in the `all_of(x)`, it would still report the full `c(all_of(x), y)` as the place where the rename happened, but I think that is ok, it is hard to be more precise. I still think this would be a big improvement.


---

I think adding the argument name is nice, but maybe not quite as good as it could be on its own. For example this dplyr error from https://github.com/tidyverse/dplyr/issues/6745 would have been:

```r
#> Error in `rename_with()`:
#> ! Can't rename variables in this context.
#> i `.cols` can't be renamed.
```

But the user typed in `mtcars %>% rename_with(toupper, all_of(v))` so `.cols` isn't immediately meaningful to them. But this:

```r
#> Error in `rename_with()`:
#> ! Can't rename variables in this context.
#> i Rename detected in `.cols`, with expression `all_of(v)`.
```

is probably enough for them to figure out where to look next, because it refers to something in their code

WDYT @lionel- ?

lionel-: I like it!

lionel-:
Closing this since it was split into new PRs.

R/eval-walk.R
lionel-: I think we could use the "In argument" pattern for consistency.

```r
#> Error in `rename_with()`:
#> ! Can't rename variables in this tidyselect context.
#> i In argument: `.cols = all_of(v)`.
```

PR #359, opened 2024-10-23
Support `error_arg` for `allow_rename = FALSE` as well

olivroy:
For #336, follow-up to #350

### DIFF
--- a/R/eval-walk.R
+++ b/R/eval-walk.R
@@ -110,11 +110,22 @@ ensure_named <- function(pos,
   check_empty(pos, allow_empty, error_arg, call = call)
 
   if (!allow_rename && any(names2(pos) != "")) {
-    cli::cli_abort(
-      "Can't rename variables in this context.",
-      class = "tidyselect:::error_disallowed_rename",
-      call = call
-    )
+    if (is.null(error_arg)) {
+      cli::cli_abort(
+        "Can't rename variables in this context.",
+        class = "tidyselect:::error_disallowed_rename",
+        call = call
+      )
+    } else {
+      cli::cli_abort(c(
+        "Can't rename variables in this context.",
+        i = "{.arg {error_arg}} can't be renamed."
+        ),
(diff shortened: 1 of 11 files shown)


R/eval-walk.R
```diff
@@ -110,11 +110,22 @@ ensure_named <- function(pos,
...
+        call = call
+      )
+    } else {
+      cli::cli_abort(c(
+        "Can't rename variables in this context.",
+        i = "{.arg {error_arg}} can't be renamed."
+        ),
+        class = "tidyselect:::error_disallowed_rename",
+        call = call
+      )
```
DavisVaughan, line 126: Maybe construct the `message` inside the if branch, but then fall through to a single call to `cli_abort()` so we dont duplicate the class name? (that feels like a place where we could make a typo to get out of sync). Maybe like

```r
message <- "Can't rename variables in this context."

if (!is.null(error_arg)) {
  message <- c(message, i = cli::format_inline("{.arg {error_arg}} can't be renamed."))
}

cli::cli_abort(
  message,
  class = "tidyselect:::error_disallowed_rename",
  call = call
)
```

(Or maybe apply this general idea to the final error message that incorporates the `expr`)

olivroy: Good idea! I caught a copy paste typo while working on this.

I applied a similar logic to another duplicated error class I created

lionel-: In that case it's good practice to still interpolate `message` in a string like this:

```r
cli::cli_abort(
  "{message}",
  class = "tidyselect:::error_disallowed_rename",
  call = call
)
```

This prevents unexpected interpolation/interpretation of components of messages, esp. if it comes from external sources (little bobby tables).

lionel-: oh but here it's not as straightforward because we're adding an element to the vector... I'm just going to revert this change then.

lionel-:
Thanks for all these!

PR #361, opened 2024-10-23
Avoid nested `expect_error()` in snapshot tests

olivroy:
I guess the goal was to have the class of the error.

This PR uses the `cnd_class` argument from `expect_snapshot()` instead.

I decided to use 

```r
expect_snapshot(error = TRUE, cnd_class = TRUE, {
  code
})
```

to avoid having a large diff if it is decided the condition class is no longer important in certain tests.

### DIFF
--- a/tests/testthat/_snaps/vars-pull.md
+++ b/tests/testthat/_snaps/vars-pull.md
@@ -79,22 +79,19 @@
       Error in `f()`:
       ! `var` is absent but must be supplied.
 
-# vars_pull() has informative errors
+# vars_pull() are base errors
 
     Code
-      # # vars_pull() instruments base errors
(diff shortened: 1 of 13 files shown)


lionel-:
> I guess the goal was to have the class of the error.

I think it was also to see the error call at a time where testthat would not show it.

tests/testthat/_snaps/vars-pull.md
```diff
@@ -79,22 +79,19 @@
       Error in `f()`:
       ! `var` is absent but must be supplied.
 
-# vars_pull() has informative errors
+# vars_pull() are base errors
```
lionel-, line 82: I don't understand this new title

PR #362, opened 2024-10-24
Add `cross` argument to `num_range()`

### LINKED ISSUE #355: `num_range()` docs could be more explicit about the fact that `prefix`, `range`, and `suffix` are all recycled
See https://github.com/tidyverse/tidyr/issues/1552, where the user was expecting `prefix` and `range` to generate an expand-grid style alternating combination of options, which is also reasonable to expect without more documentation about this

lionel-:
And use tidyverse recycling rules if `FALSE`

Closes #355.


R/helpers-pattern.R
```diff
@@ -174,14 +174,29 @@ matches <- function(match,
 #' @param range A sequence of integers, like `1:5`.
 #' @param width Optionally, the "width" of the numeric range. For example,
 #'   a range of 2 gives "01", a range of three "001", etc.
+#' @inheritParams rlang::args_dots_empty
+#' @param cross Whether to take the cartesian product of `prefix`, `range`, and `suffix`.
+#'   If `FALSE`, the default, these arguments are recycled using tidyverse rules.
```
DavisVaughan, line 179:
```suggestion
#'   If `FALSE`, the default, these arguments are recycled using [tidyverse rules][vctrs::vector_recycling_rules].
```

+ redocument

I think that's how the crosslink is specified

Historically people in dplyr have asked me to cross link to the recycling rules because they are like "well, what are they?"

R/helpers-pattern.R
```diff
@@ -174,14 +174,29 @@ matches <- function(match,
...
                       suffix = "",
                       width = NULL,
+                      ...,
+                      cross = FALSE,
                       vars = NULL) {
+  check_dots_empty()
   vars <- vars %||% peek_vars(fn = "num_range")
 
+  if (cross) {
+    args <- vctrs::vec_expand_grid(prefix = prefix, range = range, suffix = suffix)
```
DavisVaughan, line 192:
```suggestion
    args <- vctrs::vec_expand_grid(prefix = prefix, range = range, suffix = suffix, .error_call = current_env())
```

Probably want this, it defaults to `.error_call = current_env()` because I think we determined its a tool people may use at top level

R/helpers-pattern.R
```diff
@@ -174,14 +174,29 @@ matches <- function(match,
...
+  } else {
+    args <- vctrs::vec_recycle_common(prefix = prefix, range = range, suffix = suffix)
+  }
+  prefix <- args$prefix
+  range <- args$range
+  suffix <- args$suffix
+
   if (!is_null(width)) {
```
DavisVaughan, line 200: Is `width` supposed to be length 1? Right now nothing enforces that and in theory someone could provide a `width` the same size as `range`, only for `range` to get cartesian-producted to a different size

R/helpers-pattern.R
```diff
@@ -174,14 +174,29 @@ matches <- function(match,
...
+#' @inheritParams rlang::args_dots_empty
+#' @param cross Whether to take the cartesian product of `prefix`, `range`, and `suffix`.
+#'   If `FALSE`, the default, these arguments are recycled using tidyverse rules.
 #' @export
 num_range <- function(prefix,
                       range,
                       suffix = "",
                       width = NULL,
+                      ...,
+                      cross = FALSE,
```
DavisVaughan, line 186: `check_bool(cross)` if you want

DavisVaughan (approved):

R/helpers-pattern.R
lionel-: oh, so confusing that other `vec_` tools use the caller env

R/helpers-pattern.R
lionel-: I added a scalar number check for simplicity.

PR #364, opened 2024-10-24
Deprecate `.data` in `pull()` too

### LINKED ISSUE #335: Should `pull()` also give a deprecation warning about `.data$var`?
I was surprised when upgrading to `tidyselect` v1.2.0 that `pull()` doesn't also warn about using the `.data$var` setup.

```
> data.frame(x = 1) %>% pull(.data$x)
[1] 1
> data.frame(x = 1) %>% select(.data$x)
Warning: Use of .data in tidyselect expressions was deprecated in tidyselect 1.2.0.
ℹ Please use `"x"` instead of `.data$x`
This warning is displayed once every 8 hours.
Call `lifecycle::last_lifecycle_warnings()` to see where this warning was generated.
  x
1 1
```

Happy to provide sessionInfo if it's helpful

lionel-:
Closes #335.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # tidyselect (development version)
 
+* `vars_pull()` now also warns when using `.data` (#335). Please
+  use string-quotation programmatic usage, consistently with other
+  tidyselect contexts.
+
 * `num_range()` now recycles its arguments using tidyverse rules (#355).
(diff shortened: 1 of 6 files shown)


DavisVaughan (approved):
NEWS bullet? Kind of optional here i guess since you are just catching `vars_pull()` up

PR #369, opened 2024-10-28
fixup error msg

olivroy:
Also remove duplicate news items (it is already on line 15)

Noted that `with_rename_error()` is no longer needed as using `eval_select()` with `error_arg` does the job (better?)


lionel-:
Thanks!

